PR #2, opened 2018-11-16
Mepts 22 tx pvls indicator

ningosi:
@nthfloor here is my initial work for TX PVLS report for review


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxPvlsDataset.java
```diff
@@ -0,0 +1,211 @@
...
+		ColumnParameters fftyAndAboveF = new ColumnParameters(">=50F", ">=50 years female", "gender=F|age=>49");// 26
+		ColumnParameters fftyAndAboveT = new ColumnParameters(">=50T", ">=50 years patients", "age=>49");// 27
+		
+		return Arrays.asList(tenTo14M, tenTo14F, tenTo14T, fftnTo19M, fftnTo19F, fftnTo19T, twtyTo24M, twtyTo24F, twtyTo24T,
+		    twty5To29M, twty5To29F, twty5To29T, thtyTo34M, thtyTo34F, thtyTo34T, thty5To39M, thty5To39F, thty5To39T,
+		    ftyTo44M, ftyTo44F, ftyTo44T, fty5To49M, fty5To49F, fty5To49T, fftyAndAboveM, fftyAndAboveF, fftyAndAboveT);
+		
+	}
+	
+	private List<String> adultColumns() {
```
nthfloor: It doesn't make sense to be defining an array with columnNames, when we already have a class which is meant to represent the column. Lets maybe move this to the ColumnProperties class.

api/src/main/java/org/openmrs/module/eptsreports/reporting/utils/EptsReportUtils.java
```diff
@@ -72,4 +80,40 @@ public static void setupReportDefinition(ReportManager reportManager) {
...
+	 * @param baseLabel the base column label
+	 * @param indicator the indicator
+	 * @param columns the column parameters
+	 * @param columnNames the column names
+	 */
+	public static void addRow(CohortIndicatorDataSetDefinition cohortDsd, String baseName, String baseLabel,
+	        Mapped<CohortIndicator> indicator, List<ColumnParameters> columns, List<String> columnNames) {
+		int c = 0;
+		for (ColumnParameters column : columns) {
+			String name = baseName + "-" + (columnNames != null ? columnNames.get(c++) : column.getName());
```
nthfloor: This if statement is not useful at the moment since columnNames is hardcoded with values and so is never null. 

I think we need to rework this bit of logic - it doesn't make sense to be defining an array with columnNames, when we already have a class which is meant to represent the column.

api/src/main/java/org/openmrs/module/eptsreports/ColumnParameters.java
```diff
@@ -0,0 +1,66 @@
...
+ * under the License.
+ *
+ * Copyright (C) OpenMRS, LLC.  All Rights Reserved.
+ */
+package org.openmrs.module.eptsreports;
+
+/**
+ * Show the parameters that can be passed to a column for disaggregation
+ */
+public class ColumnParameters {
```
nthfloor: Seems to be specific to datasets, correct?

Can we maybe look at adding this logic to a `BaseDataSet.java` file so that the logic is shared for all datasets? Could be an inner class or a global array with a function for adding columns to a `dataSetColumns` global array.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/indicators/TbIndicators.java
```diff
@@ -22,7 +23,7 @@
 public class TbIndicators extends BaseIndicators {
 	
 	public CohortIndicator tuberculosePatientNewlyInitiatingARTIndicator(CohortDefinition cohortDefinition) {
-		return newCohortIndicator("tuberculosePatientNewlyInitiatingARTIndicator", cohortDefinition,
-		    ParameterizableUtil.createParameterMappings("onOrAfter=${startDate},onOrBefore=${endDate}"));
```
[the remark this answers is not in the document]
ningosi: No benefits at all, but just wanted to uniformly use a method that has already those encapsulated in EptsReportUtils.map which can be used a cross

api/src/main/java/org/openmrs/module/eptsreports/reporting/utils/EptsReportUtils.java
```diff
@@ -72,4 +80,40 @@ public static void setupReportDefinition(ReportManager reportManager) {
...
+	 * Adds a row to a dataset based on an indicator and a list of column parameters
+	 * 
+	 * @param cohortDsd the dataset
+	 * @param baseName the base columm name
+	 * @param baseLabel the base column label
+	 * @param indicator the indicator
+	 * @param columns the column parameters
+	 * @param columnNames the column names
+	 */
+	public static void addRow(CohortIndicatorDataSetDefinition cohortDsd, String baseName, String baseLabel,
```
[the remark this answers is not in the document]
ningosi: Yes, that is only used in the Datasets, I will change to your suggestion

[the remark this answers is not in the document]
ningosi: Done

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxPvlsDataset.java
```diff
@@ -0,0 +1,211 @@
...
+		
+		// constructing the first row of pregnant and breast feeding mothers
+		dsd.addColumn("1N", "Pregnant Women - Numerator", EptsReportUtils.map(hivIndicators.cohortIndicator("pregnant",
+		    ccq.pregnantWomenAndHasSuppressedViralLoadInTheLast12MonthsNumerator(), mappings), mappings), "");
+		
+		dsd.addColumn("1D", "Pregnant Women - Denominator", EptsReportUtils.map(hivIndicators.cohortIndicator("pregnant",
+		    ccq.pregnantWomenAndHasViralLoadInTheLast12MonthsDenominator(), mappings), mappings), "");
+		
+		dsd.addColumn("2N", "Breastfeeding - Women Numerator",
+		    EptsReportUtils.map(hivIndicators.cohortIndicator("breastfeeding",
```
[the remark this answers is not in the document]
ningosi: Done

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxPvlsDataset.java
```diff
@@ -0,0 +1,211 @@
...
+		    "");
+		
+		dsd.addColumn("0D", "Total patients with Viral load - Denominator",
+		    EptsReportUtils.map(
+		        hivIndicators.cohortIndicator("suppressed", sqlCohortQueries.getPatientsViralLoadWithin12Months(), mappings),
+		        mappings),
+		    "");
+		
+		// constructing the first row of pregnant and breast feeding mothers
+		dsd.addColumn("1N", "Pregnant Women - Numerator", EptsReportUtils.map(hivIndicators.cohortIndicator("pregnant",
```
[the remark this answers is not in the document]
ningosi: Refactored

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/indicators/HivIndicators.java
```diff
@@ -13,31 +13,72 @@
...
+		        "startDate=${startDate},endDate=${endDate},location={location}"));
+	}
+	
+	/**
+	 * Generic indicator that takes a cohortDefinition and aggregate it into numbers Probably be moved
+	 * to a common file for reuse
+	 * 
+	 * @return CohortIndicator
+	 */
+	public CohortIndicator cohortIndicator(String name, CohortDefinition cohort, String mappings) {
```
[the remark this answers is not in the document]
ningosi: Removed

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/SqlCohortQueries.java
```diff
@@ -125,4 +139,168 @@ public SqlCohortDefinition getPatientsNotifiedToBeOnTbTreatment() {
...
+	
+	/**
+	 *
+	 */
+	public CohortDefinition patientsWhoGaveBirthTwoYearsAgo() {
+		SqlCohortDefinition cd = new SqlCohortDefinition();
+		cd.setName("patientsWhoGaveBirthTwoYearsAgo");
+		cd.addParameter(new Parameter("startDate", "Start Date", Date.class));
+		cd.addParameter(new Parameter("location", "Location", Location.class));
+		cd.setQuery(BreastfeedingQueries.getPatientsWhoGaveBirthTwoYearsAgo());
```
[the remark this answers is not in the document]
ningosi: Refactored the sql string in a separate file for the TX_PVLS dataset

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxPvlsDataset.java
```diff
@@ -0,0 +1,211 @@
...
+		    ReportingConstants.LOCATION_PARAMETER);
+	}
+	
+	// build dimensions specific for this data set
+	/**
+	 * Age range dimension 10-14, 15-19, 20-24, 25-29, 30-34, 35-39, 40-44, 45-49, >=50
+	 * 
+	 * @return {@link org.openmrs.module.reporting.indicator.dimension.CohortDimension}
+	 */
+	private CohortDefinitionDimension age() {
```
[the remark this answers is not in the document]
ningosi: Moved to dimensions class

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/SqlCohortQueries.java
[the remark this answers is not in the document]
nthfloor: ok cool, then please use more descriptive parameter names in the query classes. `c1` is not clear what it is for, rather use something like `pregnantConcept`

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxPvlsDataset.java
ningosi: Hi @nthfloor I did not understand this, you mean I create another class called **ColumnProperties** and put this logic in? please shed more light

nthfloor: Sorry, this is a typo, I meant to say *ColumnParameters* class.

api/src/main/java/org/openmrs/module/eptsreports/ColumnParameters.java
ningosi: Agreed. I will creat a nested class which is a nicer way of logically grouping classes that are only used in one place, increases encapsulation and It can lead to more readable and maintainable code. Thanks for this point @nthfloor

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxPvlsDataset.java
ningosi: Since this is determined by the number of columns that our addRow method will display its data, it will vary for any other dataset. I will include them in the method. We can leave it as it is for now as we find a nicer way of including those into the nested columnParameter class

nthfloor: Yes I realise the number of columns is determined by the dataset, so why not use the length of the array returned by `adultsDisagregation()` to determine the number of columns? 

You can add the column name to the *ColumnParameters* class and reference that in the `addRow()` function. Right now there are two arrays being passed into `addRow()` which represent the list of columns for that dataset - why not combine?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/BreastfeedingQueries.java
```diff
@@ -0,0 +1,39 @@
...
+		        + " inner join patient_state ps on pg.patient_program_id=ps.patient_program_id"
+		        + " where pg.voided=0 and ps.voided=0 and p.voided=0 and"
+		        + " pg.program_id=8 and ps.state=27 and ps.end_date is null and"
+		        + " ps.start_date between date_add(:startDate, interval -2 year) and date_add(:startDate, interval -1 day) and location_id=:location";
+	}
+	
+	/**
+	 * GRAVIDAS INSCRITAS NO SERVIÇO TARV
+	 */
+	public static String getPregnantWhileOnArt(int concept1, int concept2, int concept3, int concept4, int enc1, int enc2,
```
nthfloor: @ningosi Should this query maybe be in `PregnantQueries.java` file?

Also can we please use a more descriptive parameter name for concept1, concept2, etc... similar to how you did it in `getWeeksPregnantOnInitialOrFollowUpConsultations()` for example.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxPvlsDataset.java
ningosi: Will apply as suggested. Done

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/BreastfeedingQueries.java
ningosi: I will move that to respective file and provide descriptive name parameters

ningosi:
@nthfloor you can give it another round of review. Thanks

api/src/main/java/org/openmrs/module/eptsreports/reporting/utils/EptsReportUtils.java
ningosi: I think this covered too. columnNames have been replaced by a property from ColumnParameters class.

nthfloor:
@ningosi As discussed on slack - I have had a quick review and here are my thoughts:
• I still have concerns about the cohorts classes and how they have been organised. We have still landed into the trap where the CompositionQueries file is large and rapidly getting bigger making it hard to maintain.
• But to avoid massive merge conflicts I'm suggesting we leave as is and focus on verifying the queries, merge into master, then go back and reorganise properly.
• At the moment I am leaning towards grouping general queries (the ones that can form the basis for other queries, such as `generalSqlQuery`), keep the age and gender separate and then group the remaining composition queries by the program (such as one for ART, one for Pregnancies, Breastfeeding, etc...).

api/src/main/java/org/openmrs/module/eptsreports/reporting/utils/EptsReportUtils.java
```diff
@@ -72,4 +80,40 @@ public static void setupReportDefinition(ReportManager reportManager) {
...
 	
+	/**
+	 * @param parameterizable
+	 * @param mappings
+	 * @param <T>
+	 * @return
+	 */
+	public static <T extends Parameterizable> Mapped<T> map(T parameterizable, String mappings) {
+		if (parameterizable == null) {
+			throw new NullPointerException("Programming error: missing parameterizable");
```
rubailly: Would be better to use an [IllegalArgumentException](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html) here and add message like "parameterizable cannot be null". A NPE in this case is misleading as it'll make you think the caller "dereferenced" a null if you see it in a stack trace.

rubailly: Would be better to use an [IllegalArgumentException](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html) here and add message like "parameterizable cannot be null". A NPE in this case is misleading as it'll make you think the caller "dereferenced" a null if you see it in a stack trace.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/BaseDataSet.java
```diff
@@ -36,17 +35,20 @@ protected void addRow(CohortIndicatorDataSetDefinition cohortDsd, String baseNam
...
 		/**
 		 * Default constructor
 		 * 
 		 * @param name the name
 		 * @param label the label
 		 * @param dimensions the dimension parameters
 		 */
-		public ColumnParameters(String name, String label, String dimensions) {
+		public ColumnParameters(String name, String label, String dimensions, String column) {
 			this.name = name;
```
rubailly: Do we still need the "name" property?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxPvlsDataset.java
```diff
@@ -178,12 +132,79 @@ private CohortDefinitionDimension age() {
...
+	private List<ColumnParameters> infants() {
+		// column parameters for children under 1
+		ColumnParameters under1M = new ColumnParameters("<1M", "<1-Male", "gender=M|age=<1");
+		ColumnParameters under1F = new ColumnParameters("<1F", "<1-Female", "gender=F|age=<1");
+		ColumnParameters under1T = new ColumnParameters("<1T", "<1-Total", "age=<1");
+
+		return Arrays.asList(under1M, under1F, under1T);
+	}
+
+	private List<ColumnParameters> children() {
```
rubailly: Would be great to rename this to "childrenDisagregation"since we also have the one for adults name "adultsDisagregation"

api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/SetupTxPvls.java
```diff
@@ -0,0 +1,89 @@
...
+		parameters.add(ReportingConstants.START_DATE_PARAMETER);
+		parameters.add(ReportingConstants.END_DATE_PARAMETER);
+		return parameters;
+	}
+	
+	@Override
+	public List<ReportDesign> constructReportDesigns(ReportDefinition reportDefinition) {
+		ReportDesign reportDesign = null;
+		try {
+			reportDesign = createRowPerPatientXlsOverviewReportDesign(reportDefinition, "TxPvls.xls", "TxPvls.xls_", null);
```
rubailly: #createRowPerPatientXlsOverviewReportDesign will now take a string parameter to initialize the report design uuid. The call will look like this:
_createXlsReportDesign(reportDefinition, "TXCURR.xls", "TXCURR.xls_", getExcelDesignUuid(), null);_

This is from our previous discussion where @ningosi  mentioned that the design resource was not being assigned the correct uuid. The proper way to make this change would have probably been to add an overload for the method and deprecate this one, but as we are in the initial development phase of the module and the method is not used in many places, I just replaced it with this one which takes the uuid parameter and [set it](https://github.com/esaude/openmrs-module-eptsreports/blob/MEPTS-23/api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/manager/EptsDataExportManager.java#L99) as the design resource uuid. @nthfloor -You'll get merger conflicts and you can refer to this comments.

api/src/main/java/org/openmrs/module/eptsreports/reporting/utils/EptsReportUtils.java
ningosi: Mhh, Ok. I will try improve on that

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/BaseDataSet.java
ningosi: That is how I would Identify my column uniquely if I would need to reference by name

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxPvlsDataset.java
ningosi: Ok, though those are specific to this data set though renaming it will be clear than it is now

api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/SetupTxPvls.java
ningosi: Sure, after merge, we will revert to common used standard approach

api/src/main/java/org/openmrs/module/eptsreports/reporting/utils/EptsReportUtils.java
ningosi: @rubailly This is fixed, you can give @nthfloor  a node to merge

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ViralLoadQueries.java
```diff
@@ -20,33 +20,16 @@
...
 	 */
-	public static String getPatientsHavingViralLoadInLast12Months(int labEncounter, int adultSegEncounter, int paedEncounter,
-	        int vlConceptQuestion) {
-		return "SELECT p.patient_id FROM  patient p INNER JOIN encounter e ON p.patient_id=e.patient_id INNER JOIN"
-		        + " obs o ON e.encounter_id=o.encounter_id WHERE p.voided=0 AND e.voided=0 AND o.voided=0 AND"
-		        + " e.encounter_type IN (" + labEncounter + "," + adultSegEncounter + "," + paedEncounter
-		        + ") AND o.concept_id=" + vlConceptQuestion + " AND o.value_numeric IS NOT NULL AND"
-		        + " e.encounter_datetime BETWEEN date_add(:endDate, interval -12 MONTH) AND :endDate AND"
-		        + " e.location_id=:location";
+	public static String getPatientsHavingViralLoadInLast12Months(int labEncounter, int adultSegEncounter, int paedEncounter, int vlConceptQuestion) {
```
nthfloor: I actually prefer these SQL queries to be written over multiple lines - makes it easier to read. Something like this:

SELECT ...
FROM ...
JOIN ...
WHERE ...
GROUP BY ...

etc.

But I'll merge for now and we can include this in the technical debt review we will do next sprint

nthfloor:
@ningosi just ran the PVLS report and got the following error: https://hastebin.com/tiyupahasu.sql

Parameters:
Start date: | 01/11/2018
End date: | 27/11/2018
Location: | Spring(first one)

Please take a look.

nthfloor:
@rubailly Tx_New does not have a *Location* parameter like the other two reports. Please add this.
screenshot from 2018-11-28 12-57-17

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ViralLoadQueries.java
ningosi: I agree, we will restructure that

nthfloor:
@ynurmahomed No, rather apply the change to this branch and then when Tx_New is working again we can merge the whole branch into *master*

Cross merging betwn branches is fine when you need something that is being reviewed but are sure that it will be merged before you are done. Otherwise best avoid it when possible, you risk mixing too much work that is still in progress that will compromise this branch.

An example is if for whatever reason we need to rollback this pull request, if we have merged your branch into this one then we will have removed both sets of changes. Best to keep them isolated as possible.

PR #7, opened 2018-12-02
Tx new tx curr report queries fix

rubailly:
TX_NEW has been fixed.
TX_CURR is generating numbers that are slightly different from what we have in the expected results document that Pinki shared, but my first guess is that the issue is in the way they combined the sql queries.  We'll have to check with them again about that.


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxCurrDataset.java
```diff
@@ -81,7 +79,7 @@ public CohortIndicatorDataSetDefinition constructTxNewDatset(List<Parameter> par
 		 * Looks for patients enrolled in ART program (program 2=SERVICO TARV -
 		 * TRATAMENTO) before or on end date
 		 */
-		CohortDefinition inARTProgramDuringTimePeriod = sqlCohortQueries.getPatientsinARTProgramDuringTimePeriod();
+		CohortDefinition enrolledBeforeEndDate = inprogramCohortQueries.createInProgramParameterizableByDateAndLocation("InARTProgram", hivMetadata.getARTProgram(), Arrays.asList("onOrBefore"), "location");
```
nthfloor: @rubailly is this cohort meant to only have one date parameter? `onOrBefore`

Aren't they all meant to have two `onOrBefore` and `onOrAfter`?

api/src/main/java/org/openmrs/module/eptsreports/metadata/CommonMetadata.java
```diff
@@ -38,6 +38,8 @@
 	
 	private String RETURN_VISIT_DATE = "e1dae630-1d5f-11e0-b929-000c29ad1d07";
 	
+	private String TRANSFER_FROM_OTHER_FACILITY = "e1da7d3a-1d5f-11e0-b929-000c29ad1d07";
```
nthfloor: Where is this used @rubailly ?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/SqlCohortQueries.java
```diff
@@ -267,14 +280,69 @@ public CohortDefinition getDeadPersons() {
...
 	
+	// Start of TX_CURR
+	
+	// Looks for patients registered as START DRUGS (answer to question 1255 = ARV
+	// PLAN is 1256 = START DRUGS) in the first drug pickup (encounter type
+	// 18=S.TARV: FARMACIA) or follow up consultation for adults and children
+	// (encounter types 6=S.TARV: ADULTO SEGUIMENTO and 9=S.TARV: PEDIATRIA
+	// SEGUIMENTO) before or on end date
+	@DocumentedDefinition(value = "patientWithSTARTDRUGSObs")
+	public CohortDefinition getPatientWithSTARTDRUGSObsBeforeOrOnEndDate() {
```
nthfloor: @rubailly why have you created these differently for TxCurr? Why only measuring `OnOrBefore` instead of before and after?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxCurrDataset.java
rubailly-cure: Because it is tx_curr (..i.e they are looking for patients who are currently in the program), we only need the endDate parameter. You want just to check if the patient is still in the program at endDate

api/src/main/java/org/openmrs/module/eptsreports/metadata/CommonMetadata.java
rubailly-cure: [Here](https://github.com/esaude/openmrs-module-eptsreports/blob/TxNew-TxCurr-report-queries-fix/api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/SqlCohortQueries.java#L294), in the #hivMetadata.getTransferFromOtherFacilityConcept() method. The specs says we want patients for whom the answer to the  question "ARV Plan" is "Start drugs" **OR** "Transfer from other facility"

nthfloor: See it, not sure how I missed it before.

Thanks

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/SqlCohortQueries.java
rubailly-cure: Same as the first comment, it is because we want patients who are **currently** documented as "started on ART". they may have been started on ART a long time ago, we just want to know the current status, which is why we only use the endDate parameter.

OpenMRSFormatter.xml
```diff
@@ -81,7 +81,7 @@
         <setting id="org.eclipse.jdt.core.formatter.insert_space_before_opening_angle_bracket_in_type_parameters" value="do not insert"/>
         <setting id="org.eclipse.jdt.core.formatter.insert_new_line_in_empty_type_declaration" value="do not insert"/>
         <setting id="org.eclipse.jdt.core.formatter.comment.clear_blank_lines_in_block_comment" value="true"/>
-        <setting id="org.eclipse.jdt.core.formatter.lineSplit" value="1250"/>
+        <setting id="org.eclipse.jdt.core.formatter.lineSplit" value="135"/>
```
nthfloor: Shorted this so that code is more readable on smaller screens

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/CompositionCohortQueries.java
```diff
@@ -21,7 +21,6 @@
 import org.openmrs.module.eptsreports.reporting.utils.EptsReportUtils;
```
nthfloor: I still want to split these queries out into the relevant classes...

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxNewCohortQueries.java
```diff
@@ -0,0 +1,204 @@
...
+		
+		TxNewForAge.getSearches().put("4", new Mapped<CohortDefinition>(patientsWithDrugPickUpEncounters,
+		        ParameterizableUtil.createParameterMappings("onOrAfter=${onOrAfter},onOrBefore=${onOrBefore},location=${location}")));
+		
+		TxNewForAge.getSearches().put("5", new Mapped<CohortDefinition>(transferredFromOtherHealthFacility,
+		        ParameterizableUtil.createParameterMappings("onOrAfter=${onOrAfter},onOrBefore=${onOrBefore},location=${location}")));
+		
+		String compositionString = "((1 AND 2 AND 3 AND 4) AND (NOT 5))";
+		
+		if (AgeCohort != null) {
```
nthfloor: I don't like this, I think we can do this better, but this is the first step towards improving how the cohorts are grouped.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxNewCohortQueries.java
```diff
@@ -0,0 +1,217 @@
...
+	// Looks for patients enrolled in ART program (program 2=SERVICO TARV -
+	// TRATAMENTO) before or on end date
+	@DocumentedDefinition(value = "inARTProgramDuringTimePeriod")
+	public CohortDefinition getPatientsinARTProgramDuringTimePeriod() {
+		SqlCohortDefinition inARTProgramDuringTimePeriod = new SqlCohortDefinition();
+		inARTProgramDuringTimePeriod.setName("inARTProgramDuringTimePeriod");
+		inARTProgramDuringTimePeriod
+		        .setQuery("select p.patient_id from patient p inner join patient_program pg on p.patient_id=pg.patient_id "
+		                + "where pg.voided=0 and p.voided=0 and pg.program_id= " + hivMetadata.getARTProgram().getProgramId()
+		                + " and pg.date_enrolled <= :onOrBefore and and pg.location_id=:location");
```
ynurmahomed: Small typo here `:onOrBefore and and pg.location_id`

nthfloor: Not really a typo because that depends on the name of the parameter - in this case the parameter is called `onOrBefore`. We could change it to be `endDate`, but it won't change the query.

ningosi: It would sound correct if it is enrolledOnOrBefore and mapped to endDate in respective compositions and indicators, like it is here https://github.com/openmrs/openmrs-module-reporting/blob/master/api/src/main/java/org/openmrs/module/reporting/cohort/definition/ProgramEnrollmentCohortDefinition.java#L39

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxNewCohortQueries.java
ningosi: Sure, we can rewrite this better

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxNewCohortQueries.java
ynurmahomed: I was referring to the `and and` :sweat_smile:

nthfloor: lol gotcha :+1:

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TxNewQueries.java
```diff
@@ -32,25 +34,47 @@
...
+		// parameters are provided
+		String minAgeCondition = "1";
+		String maxAgeCondition = "1";
+		
+		if (parameters.get("minAge") != null) {
+			minAgeCondition = "timestampdiff(year,birthdate,data_inicio)>=" + parameters.get("minAge");
+		}
+		
+		if (parameters.get("maxAge") != null) {
+			maxAgeCondition = "timestampdiff(year,birthdate,data_inicio)<" + parameters.get("maxAge");
```
ynurmahomed: @nthfloor shouldn't it be `maxAgeCondition = "timestampdiff(year,birthdate,data_inicio)<=" + parameters.get("maxAge");` (with **<=**)

nthfloor: Thanks @ynurmahomed this was it.

To cater for the "<1" age cohort I have left it like this and instead updated the Agecohorts in the txNewDataset

PR #9, opened 2018-12-10
Fix TxCurr queries

edrisse:


api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
```diff
@@ -107,12 +107,12 @@ public Program getPtvEtvProgram() {
...
 	
+	private String TRANSFERED_TO_ANOTHER_FACILITY = "TRANSFERRED OUT TO ANOTHER FACILITY";
+	
 	public ProgramWorkflowState gettransferredFromOtherHealthFacilityWorkflowState() {
 		String artProgramUuid = Context.getAdministrationService()
 		        .getGlobalProperty(EptsReportConstants.GLOBAL_PROPERTY_ART_PROGRAM_UUID);
-		String transferFromOtherUuid = Context.getAdministrationService()
-		        .getGlobalProperty(EptsReportConstants.GLOBAL_PROPERTY_TRANSFER_FROM_OTHER_FACILITY_CONCEPT_UUID);
-		return getProgramWorkflowState(artProgramUuid, "2", transferFromOtherUuid);
+		return getProgramWorkflowState(artProgramUuid, "2", TRANSFERED_TO_ANOTHER_FACILITY);
```
ynurmahomed: It is expected that this method return the TRANSFER FROM OTHER FACILITY state. TRANSFERRED OUT TO ANOTHER FACILITY should be a different method.

edrisse: @ynurmahomed fixed, please check again

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/GenericCohortQueries.java
```diff
@@ -158,9 +160,13 @@ public CohortDefinition createInProgram(String name, Program program) {
 	 * @return CohortDefinition
 	 */
 	public CohortDefinition getBaseCohort() {
-		return generalSql("baseCohort",
-		    BaseQueries.getBaseCohortQuery(hivMetadata.getARVAdultInitialEncounterType().getEncounterTypeId(),
-		        hivMetadata.getARVPediatriaInitialEncounterType().getEncounterTypeId(), hivMetadata.getHIVCareProgram().getProgramId(),
-		        hivMetadata.getARTProgram().getProgramId()));
+		Map<String, String> parameters = new HashMap<>();
```
nthfloor: `new HashMap<>()` needs to have types defined - this prevents the module from loading successfully.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
```diff
@@ -228,8 +211,9 @@ public CohortDefinition getTxCurrCompositionCohort(String cohortName, CohortDefi
 		TxCurrComposition.addParameter(new Parameter("onOrBefore", "onOrBefore", Date.class));
 		TxCurrComposition.addParameter(new Parameter("location", "location", Location.class));
 		TxCurrComposition.addParameter(new Parameter("effectiveDate", "effectiveDate", Date.class));
+		TxCurrComposition.addParameter(new Parameter("locations", "location", Location.class));
 		TxCurrComposition.getSearches().put("1", new Mapped<CohortDefinition>(inARTProgramAtEndDate,
-		        ParameterizableUtil.createParameterMappings("onOrBefore=${onOrBefore},location=${location}")));
+		        ParameterizableUtil.createParameterMappings("onOrBefore=${onOrBefore},locations=${location}")));
```
nthfloor: `locations` should be singular since we are talking about just one location

api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
```diff
@@ -114,6 +114,14 @@ public Program getHIVCareProgram() {
 	
 	private String PATIENT_HAS_DIED = "PATIENT HAS DIED";
 	
+	private String TRANSFERED_OUT_TO_ANOTHER_FACILITY = "TRANSFERRED OUT TO ANOTHER FACILITY";
```
nthfloor: Do we need to have these strings kept outside of the get function? Is there a benefit for doing it this way?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/GenericCohortQueries.java
```diff
@@ -148,7 +150,7 @@ public CohortDefinition createInProgram(String name, Program program) {
 		
 		inProgram.setPrograms(programs);
 		inProgram.addParameter(new Parameter("onOrBefore", "Before Date", Date.class));
-		inProgram.addParameter(new Parameter("location", "Location", Location.class));
+		inProgram.addParameter(new Parameter("locations", "Location", Location.class));
```
nthfloor: `locations` should be singular since it is referring to one location

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/indicators/HivIndicators.java
```diff
@@ -47,24 +51,24 @@ public CohortIndicator patientEnrolledInHIVStartedARTIndicator(CohortDefinition
 	}
 	
 	public CohortIndicator patientBelow1YearEnrolledInHIVStartedARTIndicatorBeforeOrOnEndDate(CohortDefinition cohortDefinition) {
```
nthfloor: I know this was here before, but these indicators actually look the same - they are more of a generic indicator which takes in a cohort and returns an indicator. Please merge into one.

Also pls remove the duplicated parameters of `location` & `locations` - there should just be one for consistency

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/indicators/PregnantIndicators.java
```diff
@@ -18,9 +18,8 @@
 	 * @return CohortIndicator
 	 */
 	public CohortIndicator getPregnantWomenWithSuppressedViralLoadIn12Months() {
```
nthfloor: Same here - these all look the same. I think we can merge into one generic indicator and pass in the name and the cohort definition as parameters

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
edrisse: @nthfloor if we use `location` the code will not work. The parameter name must match the field on the Cohort class we are using.

nthfloor: Then please update the Cohort class

api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
edrisse: @nthfloor these are just constants (missing the `final` and `static` though), I think we can leave them as class members so we can use in other methods if needed.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
nthfloor: We need to try and unify the parameter names we are using - we have too many variations

edrisse: It's the framework class `InProgramCohortDefinition`, we can't change

nthfloor: No that is just a label we defined: have a look at L153 in the `GenericCihortQueries` file - it was named as `locations` but could have been named anything else really

edrisse: Sorry Nathan, it's not. You will not have errors but the location will not be taken into account. Please test and see.

api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
nthfloor: Are these constants being used anywhere else? They don't seem reusable to me - it is just the name of the workflow-state.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/GenericCohortQueries.java
edrisse: This has been fixed

api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
edrisse: I usually just leave constants as class members. How should I fix this, use the string directly or declare the constant inside the method?

nthfloor: Well I don't consider names as constants unless it is something I plan on reusing elsewhere. So in this case I would use the string directly

PR #15, opened 2018-12-19
MEPTS-51_TXPVLS_Changes: Added minor changes on OnArtForLessThanXmonthsCalcultion

Guimino:
getting viraloadConcept from HivMetadata  and  created method to check if objects are  not null

### DIFF
--- a/api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/pvls/OnArtForLessThanXmonthsCalcultion.java
+++ b/api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/pvls/OnArtForLessThanXmonthsCalcultion.java
@@ -1,31 +1,37 @@
 package org.openmrs.module.eptsreports.reporting.calculation.pvls;
 
+import java.util.Collection;
+import java.util.Date;
+import java.util.Map;
+
 import org.openmrs.Concept;
 import org.openmrs.Obs;
-import org.openmrs.api.context.Context;
 import org.openmrs.calculation.patient.PatientCalculationContext;
 import org.openmrs.calculation.result.CalculationResultMap;
 import org.openmrs.calculation.result.SimpleResult;
+import org.openmrs.module.eptsreports.metadata.HivMetadata;
 import org.openmrs.module.eptsreports.reporting.calculation.AbstractPatientCalculation;
 import org.openmrs.module.eptsreports.reporting.calculation.BooleanResult;
 import org.openmrs.module.eptsreports.reporting.calculation.EptsCalculations;
 import org.openmrs.module.eptsreports.reporting.utils.EptsCalculationUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
 
-import java.util.Collection;
-import java.util.Date;
-import java.util.Map;
-import java.util.Set;
-
+@Component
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/pvls/OnArtForLessThanXmonthsCalcultion.java
```diff
@@ -1,31 +1,37 @@
...
 	public CalculationResultMap evaluate(Collection<Integer> cohort, Map<String, Object> params, PatientCalculationContext context) {
 		
 		CalculationResultMap map = new CalculationResultMap();
-		Concept viralLoadConceptuuid = Context.getConceptService().getConceptByUuid("e1d6247e-1d5f-11e0-b929-000c29ad1d07");
+		Concept viralLoadConcept = hivMetadata.getHivViralLoadConcept();
+		
 		// get data inicio TARV
 		CalculationResultMap arvsInitiationDateMap = calculate(new InitialArtStartDateCalculation(), cohort, context);
-		CalculationResultMap lastVl = EptsCalculations.lastObs(viralLoadConceptuuid, cohort, context);
+		CalculationResultMap lastVl = EptsCalculations.lastObs(viralLoadConcept, cohort, context);
```
ningosi: I always get a null pointer exception when I Autowire components and use them, this made me revert to the current way am using. Did you test that and make sure the NPE is NOT being thrown?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxPvlsCohortQueries.java
```diff
@@ -52,7 +51,7 @@
 	 */
 	public CohortDefinition getPatientsWhoAreLessThan3MonthsOnArt() {
 		CalculationCohortDefinition cd = new CalculationCohortDefinition("On ART for less than 3 months",
-		        new OnArtForLessThanXmonthsCalcultion());
+		        Context.getRegisteredComponents(OnArtForLessThanXmonthsCalcultion.class).get(0));
```
ningosi: My evaluator will need new objects(instances) of the calculation being created, why would we want to go through the Context for this? any benefits for this that the previous method?

api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/pvls/OnArtForLessThanXmonthsCalcultion.java
Guimino: @ningosi , that's weird, but there is no NPE in that code, you can check. maybe you were not putting the correct annotation in the component class.

nthfloor: Works for me too, I am not getting a NPE. Agree with Guimino's comment about the annotations and we've used this approach in the other reports so should work here too.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxPvlsCohortQueries.java
Guimino: I don't think we need to  be instantiating calculation classes to the evaluator, the context will provide one instance of the OnArtForLessThanXmonthsCalcultion

PR #25, opened 2018-12-22
Disabling the independent reports

edrisse:


api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/SetupTxCurr21.java
```diff
@@ -23,6 +23,7 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+@Deprecated
```
nthfloor: Please keep this one enabled. Partners need the old spec to provide results to MISAU

PR #28, opened 2019-01-04
Format source

ynurmahomed:
Formatting files that were not originally formatted, to prevent them from being shown in git changed files


edrisse:
@psmatsinhe can we commit this? This code formatting differences are not helping

PR #30, opened 2019-01-07
Mepts 44 - Refactor indicators

ningosi:
@psmatsinhe checkout this PR for https://jembiprojects.jira.com/browse/MEPTS-44


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxNewCohortQueries.java
```diff
@@ -36,12 +29,17 @@
 import org.openmrs.module.reporting.cohort.definition.SqlCohortDefinition;
 import org.openmrs.module.reporting.common.SetComparator;
 import org.openmrs.module.reporting.definition.library.DocumentedDefinition;
-import org.openmrs.module.reporting.evaluation.parameter.Mapped;
 import org.openmrs.module.reporting.evaluation.parameter.Parameter;
-import org.openmrs.module.reporting.evaluation.parameter.ParameterizableUtil;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+import java.util.ArrayList;
```
nthfloor: Please keep the ordering of these imports to be consistent with the rest of the files in this module.

The order we have adopted is to have the most generic imports first. Here is a description: https://petroware.no/javastyle.html#Package_and_Import_Statements

ningosi: I will make them consistent

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/indicators/BreastfeedingIndicators.java
```diff
@@ -73,14 +73,4 @@ public CohortIndicator getBreastfeedingWomenWithViralLoadIn12MonthsAndNotDocumen
...
 		    "startDate=${startDate},endDate=${endDate},location=${location}"));
 	}
-	
-	/**
-	 * Breastfeeding and pregnant women on routine
-	 * 
-	 * @return CohortIndicator
-	 */
-	public CohortIndicator getBreastfeedingAndPregnantOnRoutine() {
-		return newCohortIndicator("breastfeedingWomenWithViralLoad", EptsReportUtils.map(
```
nthfloor: You'll notice that they all call `newCohortIndicator()` which means we are adding a layer which is not providing much value. 

I was thinking we actually move the `newCohortIndicator()`  calls to the dataset, which will remove the need for these files and reduce overall clutter of the module.

ningosi: Right, am reworking on those

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/indicators/HivIndicators.java
```diff
@@ -26,16 +26,6 @@
...
-		return newCohortIndicator("patientBelow1YearEnrolledInHIVStartedARTIndicator", EptsReportUtils.map(cohortDefinition,
-		    "onOrAfter=${startDate},onOrBefore=${endDate},location=${location},effectiveDate=${endDate}"));
-	}
-	
-	public CohortIndicator patientBetween1And9YearsEnrolledInHIVStartedARTIndicator(CohortDefinition cohortDefinition) {
-		return newCohortIndicator("patientBelow1YearEnrolledInHIVStartedARTIndicator", EptsReportUtils.map(cohortDefinition,
-		    "onOrAfter=${startDate},onOrBefore=${endDate},location=${location},effectiveDate=${endDate}"));
-	}
-	
 	public CohortIndicator patientInYearRangeEnrolledInHIVStartedARTIndicator(CohortDefinition cohortDefinition) {
```
nthfloor: @ningosi can we remove all these indicators and replace with calls directly to `newCohortIndicator()`?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/TxPvlsDataset.java
```diff
@@ -64,51 +64,40 @@ public DataSetDefinition constructTxPvlsDatset() {
...
-		        breastfeedingIndicators.getBreastfeedingWomenWithViralLoadIn12MonthsAndOnRoutineDenominator(), mappings), "");
+		        eptsGeneralIndicator.getIndicator("breastfeedingWomenWithViralLoad-routine denominator", EptsReportUtils.map(txPvls.getBreastfeedingWomenOnRoutineWithViralLoadResultsDenominator(), mappings) ), mappings),
+				"");
 		dsd.addColumn(
 		    "B06",
 		    "Breast feeding and NOT documented Denominator",
 		    EptsReportUtils.map(
-		        breastfeedingIndicators.getBreastfeedingWomenWithViralLoadIn12MonthsAndNotDocumentedDenominator(), mappings),
-		    "");
+		        eptsGeneralIndicator.getIndicator("breastfeedingWomenWithViralLoad-not documented denominator", EptsReportUtils.map(txPvls.getBreastfeedingWomenAndNotDocumentedWithViralLoadResultsDenominator(), mappings) ), mappings),
```
nthfloor: The second mappings parameter is not needed here. Same applies a few times in this file.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/indicators/HivIndicators.java
ningosi: Yes, I will look into those

PR #34, opened 2019-01-14
MEPTS-73: Add exclusion on TxCurr

Guimino:
Added one more exclusion of patients enrolled in artProgram but without any encounter

### DIFF
--- a/api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
+++ b/api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
@@ -1,15 +1,13 @@
 /*
- * The contents of this file are subject to the OpenMRS Public License
- * Version 1.0 (the "License"); you may not use this file except in
- * compliance with the License. You may obtain a copy of the License at
- * http://license.openmrs.org
+ * The contents of this file are subject to the OpenMRS Public License Version
+ * 1.0 (the "License"); you may not use this file except in compliance with the
+ * License. You may obtain a copy of the License at http://license.openmrs.org
  *
- * Software distributed under the License is distributed on an "AS IS"
- * basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See the
- * License for the specific language governing rights and limitations
- * under the License.
+ * Software distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for
+ * the specific language governing rights and limitations under the License.
  *
- * Copyright (C) OpenMRS, LLC.  All Rights Reserved.
+ * Copyright (C) OpenMRS, LLC. All Rights Reserved.
  */
 package org.openmrs.module.eptsreports.reporting.library.cohorts;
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
```diff
@@ -318,11 +344,16 @@ public CohortDefinition getTxCurrCompositionCohort(String cohortName, CohortDefi
...
+		    new Mapped<CohortDefinition>(patientsEnrrolledOnArtWithoutAnyConsultation, ParameterizableUtil
+		            .createParameterMappings("location=${location}")));
+		
 		String compositionString;
 		if (currentSpec) {
-			compositionString = "(1 OR 2 OR 3 OR 4) AND (NOT (5 OR ((6 OR (NOT 11)) AND (NOT (7 OR 8))))) AND (11 OR 12)";
+			compositionString = "(1 OR 2 OR 3 OR 4) AND (NOT (5 OR ((6 OR (NOT 11)) AND (NOT (7 OR 8))))) AND (11 OR 12) AND NOT 13";
 		} else {
-			compositionString = "(1 OR 2 OR 3 OR 4) AND (NOT (5 OR (6 AND (NOT (7 OR 8)))))";
+			compositionString = "(1 OR 2 OR 3 OR 4) AND (NOT (5 OR (6 AND (NOT (7 OR 8))))) AND NOT 13";
```
nthfloor: @Guimino Can this maybe be excluded from the queries in the main union part of the composition by updating/fixing those queries? So that we can reduce the size of the starting set of patients instead of adding a new condition.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
```diff
@@ -40,6 +38,11 @@
 	        + "and obs.obs_datetime = (select max(encounter.encounter_datetime) from encounter "
 	        + "where encounter.encounter_type in (%s) and encounter.patient_id = obs.person_id and encounter.location_id = obs.location_id and encounter.voided = false and encounter.encounter_datetime <= :onOrBefore) ";
 	
+	private static final String EXCLUDE_PATIENTS_ENRROLLED_BUT_WITHOUT_CONSULTATION = "select distinct(person.person_id) from patient_program "
```
nthfloor: @Guimino lets improve this query by including the check that patients have an encounter using a left join - it is more efficient and a bit cleaner way to write the query instead of using an inner select on all encounters.

Something like this:
```
select distinct(person.person_id) 
from patient_program 
	join person on person.person_id = patient_program.patient_id
	left join encounter on (encounter.patient_id=patient_program.patient_id and encounter.location_id = :location)
where patient_program.location_id = :location
	and patient_program.program_id = %s 
	and person.voided = 0 	
	and encounter.patient_id is null
```

Guimino: ok @nthfloor

Guimino: @nthfloor , this sql returns the same results as the previous query  of exclusion and this last is more performatic. I will change to use this for the exclusion

PR #35, opened 2019-01-14
MEPTS-52 Tech Debt - refactor TxNew

ynurmahomed:


ynurmahomed:
Base cohort needs to be added to PEPFAR MER 2.3 Quarterly report definition for this to work, is it OK to add?

nthfloor:
Isn't the base cohort being applied to each indicator individually?

ningosi:
@ynurmahomed  by adding that will speed up the calculations(only will affect the calculations, the cohort parameter passed in every evaluate method will look for all patients in all locations if the basecohort is NOT defined) however, each indicator have that base cohort tied to them in a composition cohort definition, there is no harm in adding it on the report setup

ningosi:
And I would say every report that uses **CalculationCohortDefiniton** to aggregate the numbers, will need base cohort added to its equivalent reports definitions

ningosi:
@edrisse before closing this ticket MEPTS-52, this PR ought to have  been merged/closed

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxNewCohortQueries.java
```diff
@@ -285,13 +291,13 @@ public CohortDefinition getTxNewUnionNumerator(CohortDefinition AgeCohort) {
 		queryParameters.put("startDrugsConcept", hivMetadata.getstartDrugsConcept().getConceptId());
 		queryParameters.put("historicalDrugsConcept", hivMetadata.gethistoricalDrugStartDateConcept().getConceptId());
 		
-		if (AgeCohort != null && AgeCohort instanceof AgeCohortDefinition) {
-			queryParameters.put("minAge", ((AgeCohortDefinition) AgeCohort).getMinAge());
-			queryParameters.put("maxAge", ((AgeCohortDefinition) AgeCohort).getMaxAge());
+		if (ageCohort instanceof AgeCohortDefinition) {
```
nthfloor: @ynurmahomed why remove the null check for the ageCohort? This could be the cause of null exceptions down the line

api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/SetupMERQuarterly.java
```diff
@@ -75,6 +80,8 @@ public ReportDefinition constructReportDefinition() {
 		reportDefinition.setName(getName());
 		reportDefinition.setDescription(getDescription());
 		reportDefinition.setParameters(getParameters());
+		reportDefinition.setBaseCohortDefinition(genericCohortQueries.getBaseCohort(),
```
nthfloor: Wasn't it confirmed with Guimino and Nicholas that adding the baseCohort here doesn't work?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxNewCohortQueries.java
ynurmahomed: Its because `instanceof` will return `false` for `null` references, the null check is not necessary. Also IntelliJ was complaining :smile:

api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/SetupMERQuarterly.java
ynurmahomed: Strange, in this case its working. I've checked the numbers. @ningosi?

nthfloor: Looks like the SqlCohortDefinition includes the baseCohort when it [evaluates](https://github.com/openmrs/openmrs-module-reporting/blob/331fe70b9a96e5855ccc3bffe18a6fcbfbceb710/api/src/main/java/org/openmrs/module/reporting/cohort/definition/evaluator/SqlCohortDefinitionEvaluator.java#L46-L68), but there might still be a bug.

@ynurmahomed please test by modifying a patient record so that it should be excluded by the baseCohort. We need to be sure about this.

ningosi: @ynurmahomed ideally, by tying a base cohort to a report definition, is supposed to filter all other cohortdefinitions tied to that report by the set of ids coming from the base cohort. For every cohort evaluator, it checks if the base cohort is set and use it to intersect to the respective cohortdefiniton being evaluated, hence adding it at report definition level in our code is supposed to work and give correct values and expected, However, we tested for the TX_Curr using that approach and we happened to get discrepancies in numbers, I am NOT sure if it was only unique to that report or NOT and that is why we included it in a cohort composition and seemed to work.

ynurmahomed: @nthfloor I've removed `SERVICO TARV - CUIDADO` program and `S.TARV: ADULTO INICIAL` encounter type from patient 1138 location 281 (egpaf DB). The baseCohort excludes the patient,  and he does not show up on TX_NEW.

PR #44, opened 2019-01-30
Mepts 87 Implement PEPFAR Early Retention 2month Report

ningosi:
@nthfloor this what I have done so far, what is pending is to finalize on the disaggreagtions, Please create time and start looking at those. This is based on https://docs.google.com/document/d/1qviUx2-PWtqQ1S7IHfj-JAOr4l6imTz9UsTpPAw7vWI/edit?ts=5c4ae2fa#


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/Eri2MonthsCohortQueries.java
```diff
@@ -0,0 +1,338 @@
...
+   *
+   * @return CohortDefinition
+   */
+  public CohortDefinition getPatientsWhoPickedUpDrugsOnTheirSecondVisit() {
+    CompositionCohortDefinition cd = new CompositionCohortDefinition();
+    cd.setName("Patients who  picked up drugs during their second visit");
+    cd.addParameter(new Parameter("endDate", "End Date", Date.class));
+    cd.addParameter(new Parameter("location", "Location", Location.class));
+    cd.addSearch(
+        "pickedDrugsAndStartedART",
```
nthfloor: I think there might be a problem here.

`getPatientsWhoDidNotPickDrugsOnTheirSecondVisit` = `A and B and not (C or DEAD or TRANS)`

Which means that this cohort is `(A and B and not (C or DEAD or TRANS)) and not (dead OR trans)`
but in the user story this should be: `A and B and C and Not (Dead or Transferred out)`

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/Eri2MonthsCohortQueries.java
```diff
@@ -0,0 +1,338 @@
...
+        "initiatedArt",
+        EptsReportUtils.map(
+            getAllPatientsWhoInitiatedArt(), "endDate=${endDate},location=${location}"));
+    cd.addSearch(
+        "pickedDrugs",
+        EptsReportUtils.map(
+            getAllPatientsWhoReturnedFor2ndConsultationOR2ndDrugsPickUpWithin33Days(),
+            "endDate=${endDate},location=${location}"));
+    cd.addSearch(
+        "dead",
```
nthfloor: Any reason for not using `getDeceasedPatients` here?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/Eri2MonthsCohortQueries.java
```diff
@@ -0,0 +1,338 @@
...
+    cd.setName("Patients who  picked up drugs during their second visit");
+    cd.addParameter(new Parameter("endDate", "End Date", Date.class));
+    cd.addParameter(new Parameter("location", "Location", Location.class));
+    cd.addSearch(
+        "pickedDrugsAndStartedART",
+        EptsReportUtils.map(
+            getAllPatientsWhoStartedArtAndPickedDrugsOnTheirNextVisit(),
+            "endDate=${endDate},location=${location}"));
+    cd.addSearch(
+        "dead",
```
nthfloor: Same here, any reason for not using `getDeceasedPatients` here?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/dimensions/EptsCommonDimension.java
```diff
@@ -205,4 +214,110 @@ public CohortDefinitionDimension maternityDimension() {
...
+        EptsReportUtils.map(
+            eri2MonthsCohortQueries.getPatientsWhoPickedUpDrugsOnTheirSecondVisit(),
+            "endDate=${endDate},location=${location}"));
+    dim.addCohortDefinition(
+        "DP",
+        EptsReportUtils.map(
+            genericCohortQueries.getDeceasedPatients(), "endDate=${endDate},location=${location}"));
+    dim.addCohortDefinition(
+        "TOP",
+        EptsReportUtils.map(
```
nthfloor: This looks like it is: `A and B and Transferred out`

but specs is looking for: `A and B and C and Transferred out`

Looks like the query C is missing.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/dimensions/EptsCommonDimension.java
```diff
@@ -205,4 +214,110 @@ public CohortDefinitionDimension maternityDimension() {
...
+            genericCohortQueries.getPatientsBasedOnPatientStates(
+                hivMetadata.getARTProgram().getProgramId(),
+                hivMetadata
+                    .getTransferredOutToAnotherHealthFacilityWorkflowState()
+                    .getProgramWorkflowStateId()),
+            "endDate=${endDate},location=${location}"));
+    dim.addCohortDefinition(
+        "STP",
```
nthfloor: Same here: This looks like it is: `A and B and Stopped/Suspended`

but specs is looking for: `A and B and C and Stopped/Suspended`

Looks like the query C is missing.

nthfloor:
Also as discussed `4.1.7 %      (Alive & Not Transferred Out and Had Consultation or Picked-up Drugs 2x/ Initiated ART minus Transferred Out)` only needs to be applied to the `All Patients` section in the template.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/Eri2MonthsCohortQueries.java
```diff
@@ -0,0 +1,338 @@
...
+        "pregnant",
+        EptsReportUtils.map(
+            getPregnantWomenRetainedOnArtFor2MonthsFromArtInitiation(),
+            "startDate=${startDate},endDate=${endDate},location=${location}"));
+    cd.addSearch(
+        "breastfeeding",
+        EptsReportUtils.map(
+            getBreastfeedingWomenRetainedOnArtFor2MonthsFromArtInitiation(),
+            "startDate=${startDate},endDate=${endDate},location=${location}"));
+    cd.setCompositionString("(all AND children) AND NOT (pregnant OR breastfeeding)");
```
nthfloor: We could maybe improve this composition because currently this: `(all AND children) AND NOT (pregnant OR breastfeeding)`

is actually: (all AND children) AND NOT ((all and pregnant) OR (all and breastfeeding))`

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/dimensions/EptsCommonDimension.java
ningosi: Yah, that is correct, will update that, the C is missing in that dimension

ningosi: Done

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/dimensions/EptsCommonDimension.java
ningosi: Done

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/Eri2MonthsCohortQueries.java
```diff
@@ -0,0 +1,338 @@
...
+  /**
+   * Get patients who have 2 months ART retention after ART initiation A
+   *
+   * @return CohortDefinition
+   */
+  public CohortDefinition getAllPatientsRetainedOnArtFor2MonthsFromArtInitiation() {
+    SqlCohortDefinition cd = new SqlCohortDefinition();
+    cd.setName("patientsRetentionFor2MonthsOnART");
+    cd.addParameter(new Parameter("endDate", "End Date", Date.class));
+    cd.addParameter(new Parameter("location" + "", "Location", Location.class));
```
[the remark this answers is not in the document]
ningosi: Thanks, how those are being inserted is a puzzle to me. Will check/correct and probably identify the source

edrisse:
> Is this composition string meant to be commented out?

Thanks for spotting that out, fixed @nthfloor

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/Eri2MonthsCohortQueries.java
```diff
@@ -212,7 +212,7 @@ public CohortDefinition getChildrenRetainedOnArtFor2MonthsFromArtInitiation() {
         "children",
         EptsReportUtils.map(
             ageCohortQueries.createXtoYAgeCohort("Children 0 - 14", 0, 14),
-            "endDate=${cohortStartDate},location=${location}"));
+            "effectiveDate=${cohortStartDate}"));
```
nthfloor: Should this be endDate instead of cohortStartDate?

edrisse: Good question. According to the spec it should be: 

> _a) Patients with birth date information registered in the system should be calculated the age of the patient at the date of initiation on ART  (birth date minus  ART initiation date ._ 
 Only with a calculation?

nthfloor: Oh dear, then it is neither cohort or reporting endDate. We will need to use `InitialArtStartDateCalculation` to calculate the initaitedART date for the patient and use that instead.

nthfloor: I think this will require a new cohort definition to use this calculation evaluator and I think it will require us to provide a location parameter for filtering the obs & encounters lookups.

edrisse: OK, will work on that

ynurmahomed: https://github.com/esaude/openmrs-module-eptsreports/blob/eeda2b7a5407ddab25a759eac87e6f972f619375/api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxNewCohortQueries.java#L300-L313 wont work?

nthfloor: Please apply this change to both indicators, thanks

nthfloor: @ynurmahomed that is based on the TxNew query, which funny enough is actually the same base query shared with these retention queries. So yes that could actually work...

If this report was not using the same base queries, then I don't think this would have worked. So in general we will need to use the initialDate calculation like was used for PVLS.

nthfloor: @ynurmahomed that cohort is based on the txNew composition which is not exactly the same composition as for these retention indicators. So i don't think it will work. 

But it could have since these retention queries share the same base query as txNew

ynurmahomed: OK. Doesn't it make sense that the ART Start date query be the same independent of indicator.

nthfloor: I agree it should be, but the txNew cohorts are still using the old queries where the age calc was baked into the txNew query. Which is why Nicholas had to extract the logic into a separate calculator.

Moving forward we should always be using the same `InitialArtStartDateCalculation` calculation to determine ART initiation date in our cohort queries.

PR #40, opened 2019-01-25
Mepts 79 updated breastfeeding and pregnant queries

ningosi:
@nthfloor here the PR for review

PR #45, opened 2019-01-31
MEPTS-83: Added unit test framework with some tests for calculations

kaweesi:


kaweesi:
@nthfloor @ningosi @edrisse criterial 2 and 3 are questionable to me, are you guys successfully running them on data and they work as were intended, i see conflict in their algorithm.

i hope routine criteria 2 is alright, i find it strange; being on ART for at-least 12 months with 2 VL with at-least 12 months away from each other when lookup is constrained to only 12 months from context now! otherwise all the rest is catered for from my side

nthfloor:
@kaweesi Looks like the logic is accurate to [the specification](https://docs.google.com/document/d/1Uiw6UA4b8_8YmkU6aYw-qFPEm18iqK7lg__tdFmClgQ/edit). If you can see a bug, lets discuss. Otherwise we need to still provide tests for these conditions.

Please also have a look at the code quality issues identified by codacy and see if you can resolve them. Thanks

kaweesi:
i find most of these checks codacy un-necessary at: https://app.codacy.com/app/esaude-epts/openmrs-module-eptsreports/pullRequest?prid=2999779 i wonder if we can disable some of them such as;
- Method names should not contain underscores
- Avoid reassigning parameters such as 'now'
- Use explicit scoping instead of the default package private level
- Avoid really long methods. (sometimes is necessary)

i can improve those but find it a conflict with some already existing standards for for-example point 1

nthfloor:
I am not sure I agree. These are international standards which make sense to me.

- Underscores in method names makes names messy and we do not use that format anywhere in this module so why break consistency now?
- Avoiding long methods is always better - helps us maintain then moving forward. Also it is good practice to design functions to a single responsibility. So if a function is getting very long, usually means it has started to cater for multiple responsbilities. There are not many cases where we absolutely have to have one large function - usually it is easy enough (and recommended!) to break them down into sub functions.  
- using explicit scoping: this is a small detail, but I don't see any problem being explicit here. We have done this everywhere else in the module.
- Avoid reassigning parameters: this is generally accepted as [bad practice](https://sbforge.org/sonar/rules/show/squid:S1226?modal=true&layout=false). We should avoid using output arguments unless necessary, they can cause [unfortunate side effects](https://refactoring.guru/remove-assignments-to-parameters). 

Regarding there being conflicts - please explain where you see conflicts?

kaweesi:
i have moved the conflict dialog onto the slack channel to get more activity on it, openmrs generally uses underscores in test methods and so as i mentioned, it's not that i don't like the checks at all, i find them unnecessary in some cases

kaweesi:
have you found out if OpenMRS disables some checks like that underscore naming!

nthfloor:
I'll have to check :+1: . But in the meantime, lets please stick to camel casing with our method names to be consistent with the rest of the module.

kaweesi:
- `Context.getRegisteredComponents` supplies @Autowire functionality thus the replacement, besides, [we experienced system slow down in msf with more autowires](https://talk.openmrs.org/t/huge-delay-on-applicationcontext-refresh-with-listeners/19399/7) and generally now encourage using `Context.getRegisteredComponents`, however that still sounds like another change that shouldn't be part of this ticket, i included it since those @autowires were returning null in unit tests except with BaseContextMockTest something to do with the way unit testing in a module is setup with BaseModuleContextSensitiveTest.
- the test environment loads test datasets into the test database, the calculations have looks of units most of which we are going to unit test separately, testing each calculator as a unit would definitely end into a kind of 'integrated test'. mocking each call was definitely be a lot more work and bad platform for more unit tests than using datasets, in this framework, the more validations around objects the more your mocks would enlarge. I used both mocks and database calls for OpenMRS framework, the mocks are populated by the database so that way i control my behaviour by real data simulation in the unit tests datasets.

looking at unit vs integrated tests in this case, i have looked into evaluate_shouldReturnFalse at https://github.com/esaude/openmrs-module-eptsreports/compare/master...ynurmahomed:test which ideally besides mocking everything is not doing any different from what we are doing in this PR, we are just using both mocks and data at the same time to avoid mocking everything. the best tests i prefer these days are TDD tests and less mocking for OpenMRS.

Any way, all said, this PR introduces a calculations test framework that would almost avail one unit test for calculate/evaluate method, it requires each calculation test be availed a cohort, calculation and a result. that unit test evaluates with some mocking and test database thus saving you time to mock on all calculations and only write unit tests for other behaviours besides basic results matching.

kaweesi:
- unit testings vs integrated testing

| unit tests  | integrated tests |
| ------------- | ------------- |
|  unit focuses on a relatively small piece of code  | done to demonstrate that different pieces of the system work together  |
| narrow in scope  |  can cover whole applications  |
| easy to write and execute  | require more effort to put together and may require resources like database instances and hardware to be allocated for them  |
| when well written,testers and users downstream should benefit from seeing fewer bugs | do a more convincing job of demonstrating the system/functionality works |
| things outside the code under test are mocked or stubbed out | May or may not mock external dependencies, at the lower end of the spectrum an integration test could be a junit test where a repository is exercised against an in-memory database |
| shouldn't have dependencies on outside systems/functionality  | toward the upper end it could be a system test verifying applications can exchange messages  |
| You decide what the unit is by looking for the smallest testable part (you well define the smallest testable unit)  | test one block/functionality of various units and how they all function together |
|Allow you to practice TDD (if you want)|Should NOT be used to practice TDD|

there're so many differences, but it gets blurry with large units with many dependencies. u may end up testing compiler paths than unit functionality.
it's recommended that unit tests Avoid database access, network requests, and the file system but generally OpenMRS unit tests have a test database, huge systems would probably mock off a dataset too

api/src/test/java/org/openmrs/module/eptsreports/reporting/calculation/EptsCalculationUtilsMock.java
```diff
@@ -0,0 +1,31 @@
+package org.openmrs.module.eptsreports.reporting.calculation;
+
+import mockit.Mock;
+import mockit.MockUp;
+import org.openmrs.Encounter;
+import org.openmrs.Obs;
+import org.openmrs.calculation.result.CalculationResultMap;
+import org.openmrs.module.eptsreports.reporting.utils.EptsCalculationUtils;
+
+public class EptsCalculationUtilsMock extends MockUp<EptsCalculationUtils> {
```
nthfloor: These helper functions can be combined with the helper function `createBasicObs` from CalculationsTestsCache into the new OpenMRSTestHelper(named for example)  class since these are specific to obs and encounters etc...

In the long run we might want to have a file per type (obs/encounter/etc... but lets wait till that becomes a need)

omod/src/main/resources/config.xml
```diff
@@ -145,16 +145,16 @@
 		<property>@MODULE_ID@.sTarvAdultoInitialAEncounterTypeUuid</property>
 		<defaultValue>e278f820-1d5f-11e0-b929-000c29ad1d07</defaultValue>
 		<description>
-			UUID for the S.TARV: ADULTO INICIAL A encounter type
+			UUID for the S.TARV: ADULTO INITIAL A encounter type
```
nthfloor: This is meant to be `INICIAL` since this is Portuguese label for the partners.

omod/src/main/resources/config.xml
```diff
@@ -145,16 +145,16 @@
...
 		</description>
 	</globalProperty>
 
 	<globalProperty>
-		<property>@MODULE_ID@.sTarvPediatriaInicialAEncounterTypeUuid
+		<property>@MODULE_ID@.sTarvPediatriaInitialAEncounterTypeUuid
 		</property>
 		<defaultValue>e278fa8c-1d5f-11e0-b929-000c29ad1d07</defaultValue>
 		<description>
-			UUID for the S.TARV: PEDIATRIA INICIAL A encounter type
```
nthfloor: This is meant to be `INICIAL` since this is Portuguese label for the partners. Please revert

kaweesi: wow, i had missed that, although language support is through localisation else we code names in Portuguese naming

api/src/test/java/org/openmrs/module/eptsreports/reporting/calculation/EptsCalculationUtilsMock.java
kaweesi: those mocks however are meant to replace their respective methods and are better kept in separate mock files as we have done, in one review up you suggested these be invoked from BasePatientCalculationTest which makes sense and so i only have createBasicObs in OpenMRSTestHelper

omod/src/main/resources/config.xml
nthfloor: I don't think localisation applies to the labels of admin fields which these are for, which is why we coded them in portuguese.

api/src/test/java/org/openmrs/module/eptsreports/reporting/calculation/EptsCalculationUtilsMock.java
nthfloor: Ah I see, my mistake. I misread what these are. I mistook these for utils of the mock classes, for some reason a read over the mock annotations! :man_facepalming: 

As you were :smile:

kaweesi:
thanks a lot @nthfloor for the reviews, it was great improving where to find things, my tests tree looks like;
```
|____test
| |____java
| | |____org
| | | |____openmrs
| | | | |____module
| | | | | |____eptsreports
| | | | | | |____api
| | | | | | | |____dao
| | | | | | | | |____EptsReportsDaoTest.java
| | | | | | | |____EptsReportsServiceTest.java
| | | | | | |____reporting
| | | | | | | |____helper
| | | | | | | | |____OpenMRSTestHelper.java
| | | | | | | | |____TestsHelper.java
| | | | | | | |____intergrated
| | | | | | | | |____calculation
| | | | | | | | | |____BasePatientCalculationTest.java
| | | | | | | | | |____CountingCalculationTest.java
| | | | | | | | | |____pvls
| | | | | | | | | | |____InitialArtStartDateCalculationTest.java
| | | | | | | | | | |____OnArtForMoreThanXmonthsCalcultionTest.java
| | | | | | | | | | |____RoutineCalculationTest.java
| | | | | | | | |____package-info.java
| | | | | | | |____mock
| | | | | | | | |____calculation
| | | | | | | | | |____EptsCalculationsMock.java
| | | | | | | | | |____EptsCalculationUtilsMock.java
| | | | | | | | |____package-info.java
| | | | | | | |____unit
| | | | | | | | |____calculation
| | | | | | | | | |____EptsCalculationsTest.java
| | | | | | | | |____package-info.java
| |____resources
| | |____calculationsTest.xml
```

nthfloor:
@kaweesi looking into the autowiring change, I think we should revert these changes back. Seems the issue with the tests returning null is [known](https://talk.openmrs.org/t/autowiring-module-defined-beans-in-the-openmrs-context-returns-null/18392). So there is a way to resolve it. 

I'd prefer if we implemented this change consistently throughout the module instead of piecemeal. So lets take note of it, I have created a Jira tasks to evaluate possible performance improvements and will note the approach of using  ` Context.getRegisteredComponents` as something to consider.

kaweesi:
do u mean defining those beans registered through the @Component annotation in applicationcontext.xml for the unit tests, u would end up doing a lot if so, i mentioned one side effect about autowires, find more evidence at: https://talk.openmrs.org/t/huge-delay-on-applicationcontext-refresh-with-listeners/19399 have you BTW tried out checking why we wrote `Context.getRegisteredComponents` @nthfloor! i would rather think it's wise evaluating the move than think of reverting it for a quick fix and extend the problem to the future. we may probably have a task to replace the rest in-case we choose to in the future and should begin advocating use of `Context.getRegisteredComponents`

nthfloor:
@kaweesi, yes we should be using applicationcontext.xml for these tests.

I have considered the performance reasons but this is not the PR to be adding those kinds of changes. I am not convinced we desperately need this performance improvement so urgently, so it is not justified to include it here. 

As I mentioned, there is a task that will consider the effect of this kind of change with comprehensive performance metrics comparing all the possible approaches. I have not seen anyone do this analysis yet, so I don't think we know for sure that this is the best approach.

kaweesi:
about the approach, responses at: https://talk.openmrs.org/t/huge-delay-on-applicationcontext-refresh-with-listeners/19399 can be helpful

nthfloor:
I have already read that thread, and we are not experiencing that sort of delay hence I do not believe we urgently need to implement this change in this PR. I'd like to see performance comparisons for our current context before justifying this sort of change.

Performance improvements are funny things, the cause of poor performance could come from a variety of sources which are often unique to that particular system, so I'd prefer a more thorough analysis to be done in order to make a more informed decision for the EPTS system.

kaweesi:
BTW, the idea of solving those NPE in this case with TestingApplicationContext.xml resource is a theoretical fix, just tried it and it didn't help, i fully think the definition of beans using annotation vs doing it within applicationContext.xml is reason for this null, definitely double definitions will result into duplicate beans. do u intend to rewrite all @components to xml equivalents in applicationContext, if those were for-example in applicationContext.xml? before i spend enough time just reverting such a functional change, would be good if u tried out something like u are trying to think @nthfloor

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/BasePatientCalculationTest.java
```diff
@@ -100,6 +100,27 @@ protected void matchOtherResultsExcept(
         otherResult.remove(i);
       }
     }
-    Assert.assertEquals(initialResult.toString(), otherResult.toString());
+    Assert.assertTrue(matchCalculationResults(initialResult, otherResult));
+  }
+
+  protected boolean matchCalculationResults(
```
kaweesi: thanks @Guimino for catching this.

nthfloor:
@kaweesi I was wrong about using applicationContext.xml. All the Spring context is already setup for us by BaseModuleContextSensitiveTest. The NPE was being caused by not consistently using Spring dependency injection. Unfortunately with Spring when you start manually instantiating dependencies instead of injecting them then it all breaks (I'm not sure why). Anyway so by updating the tests a bit to consistently use Spring's DI everything hooks nicely.

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/BasePatientCalculationTest.java
```diff
@@ -21,88 +23,101 @@
...
+	protected Map<String, Object> params = new HashMap<String, Object>();
+	
+	public abstract PatientCalculation getCalculation();
+	
+	public abstract Collection<Integer> getCohort();
+	
+	public abstract CalculationResultMap getResult();
+	
+	@Before
+	public void setUp() throws Exception {
```
kaweesi: am seeing wrong indentations @nthfloor hope u are running mvn clean install to run the formatter. i also hope these 4 last commits from your are not making me loose any of my changes!

nthfloor: I think the formatting was my git tool. For some reason it did not register the formatter changes, after refreshing I can still see outstanding changes which are are the formatter.

As for loosing any of your changes. Have a look for yourself, everything is there with the exception of the spring dependency injection adjustments.

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/pvls/InitialArtStartDateCalculationTest.java
```diff
@@ -0,0 +1,125 @@
...
+        (Location) evaluationContext.getFromCache("location"),
+        Context.getConceptService().getConcept(7777004));
+    evaluatedResult = service.evaluate(getCohort(), getCalculation(), evaluationContext);
+    Assert.assertEquals(
+        testsHelper.getDate("2008-06-01 00:00:00.0"), evaluatedResult.get(2).getValue());
+    // the rest should not be touched
+    matchOtherResultsExcept(evaluatedResult, 2);
+
+    // TODO does the results have to contain every patient from the cohort
+    // received with empty values!
```
nthfloor: What do you mean here?

InitialArtStartDateCalculation should return a date value for each patient if they are enrolled on ARV and if not then just return null.

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/pvls/RoutineCalculationTest.java
```diff
@@ -0,0 +1,128 @@
...
+        getResult().toString(),
+        service.evaluate(getCohort(), getCalculation(), params, getEvaluationContext()).toString());
+  }
+
+  /**
+   * TODO investigate.i hope routine criteria 2 is alright, i find it strange; being on ART for
+   * atleast 12 months with 2 VL with at-least 12 months away from each other when lookup is
+   * constrained to only 12 months from context now!
+   */
+  public void calculateRoutineCriteria2() {
```
nthfloor: Lets remove these functions for criteria2 & 3 since they will be covered by MEPTS-127

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/pvls/RoutineCalculationTest.java
```diff
@@ -0,0 +1,128 @@
...
+    return Context.getRegisteredComponents(RoutineCalculation.class).get(0);
+  }
+
+  @Override
+  public Collection<Integer> getCohort() {
+    return Arrays.asList(new Integer[] {2, 6, 7, 8, 999, 432});
+  }
+
+  @Override
+  public CalculationResultMap getResult() {
```
nthfloor: Please add test cases that cover permutations that cover if the patient is breastfeeding or not.

Also in order to help quickly analysing these tests and to know what scenarios these are covering, please include in the comments for each of these patients, why the routine calculation would return true/false.

Please do the same for the other test classes. We need this because the scenario is defined in the dataset being loaded into the context we should document what test scenarios have been covered.

api/src/test/java/org/openmrs/module/eptsreports/reporting/mock/calculation/EptsCalculationUtilsMock.java
```diff
@@ -0,0 +1,32 @@
...
+
+import mockit.Mock;
+import mockit.MockUp;
+import org.openmrs.Encounter;
+import org.openmrs.Obs;
+import org.openmrs.calculation.result.CalculationResultMap;
+import org.openmrs.module.eptsreports.reporting.utils.EptsCalculationUtils;
+
+/** Mocks {@link EptsCalculationUtils} with org.jmockit */
+public class EptsCalculationUtilsMock extends MockUp<EptsCalculationUtils> {
```
nthfloor: This mock class can be removed - the tests still execute successfully without them

api/src/test/java/org/openmrs/module/eptsreports/reporting/mock/calculation/EptsCalculationsMock.java
```diff
@@ -0,0 +1,199 @@
...
+import org.openmrs.calculation.patient.PatientCalculation;
+import org.openmrs.calculation.patient.PatientCalculationContext;
+import org.openmrs.calculation.result.CalculationResultMap;
+import org.openmrs.calculation.result.ListResult;
+import org.openmrs.calculation.result.SimpleResult;
+import org.openmrs.module.eptsreports.reporting.calculation.EptsCalculations;
+import org.openmrs.module.reporting.common.TimeQualifier;
+
+/** Mocks {@link EptsCalculations} with org.jmockit */
+public class EptsCalculationsMock extends MockUp<EptsCalculations> {
```
nthfloor: The only test class that depends on this is OnArtForMoreThanXmonthsCalcultionTest, so I am looking into how we can update calculationTest.xml to remove the need for this mock class.

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/pvls/InitialArtStartDateCalculationTest.java
```diff
@@ -0,0 +1,125 @@
...
+  public CalculationResultMap getResult() {
+    PatientCalculation calculation = getCalculation();
+    CalculationResultMap map = new CalculationResultMap();
+
+    // initiated ART by hiv enrolment
+    PatientCalculationContext evaluationContext = getEvaluationContext();
+    map.put(
+        2,
+        new SimpleResult(
+            new Timestamp(testsHelper.getDate("2008-08-01 00:00:00.0").getTime()),
```
nthfloor: I can see this patient has encounters and obs defined but I don't see this particular result date value defined anywhere in calculationTest.xml. So in this case where is the initialArtStartDate of 2008-08-01 being defined for this patient?

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/pvls/RoutineCalculationTest.java
nthfloor: Just to explain a bit more what I mean here - knowing that we want to know what scenario this case covers instead of just this patient's initiation date.

api/src/test/resources/calculationsTest.xml
```diff
@@ -0,0 +1,129 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
nthfloor: Lets rename this file to be specific to the PVLS calculations, something like `integrationPvlsCalculations.xml`. Better that we isolate the data context for each test set. Will make it more manageable moving forward :+1:

kaweesi: it could end up also being used for other non pvls calculations if we will have any! it takes time to setup a test dataset and its why i work of https://github.com/openmrs/openmrs-core/blob/master/api/src/test/resources/org/openmrs/include/standardTestDataset.xml until my scenarios require more data.

nthfloor: It is better that we group the necessary datasets, 2 reasons
1. If we need to modify the existing dataset for one test class it could break another test class because the scenarios are different.
2. This single dataset could become very large

In general, I think it is better to maintain vertical slices in how we layout our code. So maybe lets rather have one base dataset to contain the concepts and global properties but keep separate the encounters and obs because those will be unique to the scenarios we are trying to test in each test class.

Make sense?

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/pvls/RoutineCalculationTest.java
kaweesi: i think it's now clearer why i could have chosen breastfeeding to be another test of its own since this scenario is triggerable with setup. the basic scenario test evaluates ADULTCHILDREN. you can only test other criteria through setting it up as params and the our base evaluator handles one at a time

api/src/test/resources/calculationsTest.xml
kaweesi: but u still are better off sharing most of those encounters and obs across different scenarios, if we had no idea of reusing things in this dataset and it wasn't still small enough, it would make sense breaking it down. the encounters and obs are all tied around the same concepts we define and so i tend to be attracted to the approach of a grouped test dataset per each calculator after applying the general dataset but in this case these calculators all look tied into one another and it looks ok to me to have one dataset for them to make it easier for people to write new unit tests or even maintain existing ones

nthfloor: There is nothing stopping us from using an existing dataset in a new test, especially if the test scenarios are all covered (which means implementing new tests is still quick and easy), however if we tie everything into one dataset then it becomes a problem later on if/when we want to split them into separate files.

Better that we design with this in mind from the beginning.

Remember we want to follow the single responsiblity principle as much as possible and this applies with tests as well. Better to keep each test as isolated as possible (and within reason) from other tests so that the tests can run in a reliable and predictable environment. So using separate dataset files to cater for that particular test class makes each dataset smaller and simpler, which results in it being easier to manage and maintain as our test harness grows.

kaweesi: fully agree, i just kind of think in this case where we are testing already closely related calculations that break the single responsibility principle, it's fully sensible to have the same dataset, this can probably just act as the starting dataset for all calculations. something close to the core standardTestDataset.xml.

nthfloor: Well I can see problems will come with the encounters and obs for other calculations (perhaps not even needed in other tests) which will mean we will need to either add to this dataset or extend it with another dataset (now breaking single responsibility). Remember the obs and encounters are going to look different depending on the scenarios being tested, therefore there is no guarantee they will be shared across all test classes. 

So my suggestion is to have one base dataset for concepts and global properties that apply to all tests and separate datasets (one for PVLS, one for breastfeeding, etc...) for the encounters and obs which apply to the tests scenarios being covered in those respective classes. There is no guarantee these obs will be used across all test classes therefore they should not be in the shared base dataset

kaweesi: am not really implying that your suggestion isn't good @nthfloor, but its just not proper to reason having datasets basing on just encounters and obs, the idea of having many datasets isn't liked to me personally when testing closely related features and considering my 7 years around OpenMRS, those 3 calculators for-example depend on one another except one. having more than 6 dataset files isn't practical and i would not only consider it a less good use of dev time but also making maintenance of code harder; having to look through tens of dataset files before being able to understand closely related tests isn't good design. 
about having one base dataset for just concepts and global properties, i still consider it not perfect since all those we are tracking down for these calculations dataset are for them, loading all of them for other calculators in the base test.
the most practical would be each dataset for each test class since whether concepts or global properties or encounters or obs are closely tied together per each calculation. however it's still not good enough applying 3 datasets for a calculation that runs other calculations. 
THE main reason why i still consider our one dataset more good is it's small and easy to track down things if one is aware that standardTestDataset.xml is our base dataset

nthfloor: First off I think that is an over exaggeration that you will need to search through lots of dataset files for each test. What I am proposing that you will need to load 2 maybe 3 datasets (one shared with concepts and one specific with obs and encounters).

Let me ask you something, in the event you pick up a test class and start analysing it, will it be easier to analyse one huge dataset file or a much smaller file which has all the data relevant to that particular test?

This PR https://github.com/esaude/openmrs-module-eptsreports/pull/40/files demonstrates this nicely where the encounters and obs are different, and I think it is a good example of why having separate dataset files is  a good idea since how that test class is very easy to review and maintain.

Maybe others can also comment, @ningosi @psmatsinhe @ynurmahomed @Guimino @edrisse ?

kaweesi: obs and encounters require concepts, having a separate file for each set is unpractical,  having all concepts and then just another dataset for obs and encounters, understanding the later is tied on opening and reading through the former, but the most side effect is why load all concepts and gps in a base dataset and then break down obs and encounters into another file.

sure, understanding this datasets with not even 200 lines is easy, it saves time just tracing out matched ids and uuids in separate files.

you see! what's nice to one dev may never be similarly nice to another, am ok with other calculation tests defining extra datasets to add on the base one which exactly that commit does treating the existing one as a base dataset the way it does the standard one in openmrs, but the idea is the base one should have just enough data for shared across calculations.

any way, others could give their view about defineing all concepts and gps in a base dataset and encounters and obs in child datasets!! which would be more easy to read...

edrisse: Make the data set too centralized and you will end up with a massive data set that is hard to understand and maintain; Make it too small and you will have hardwork every time you change the data model. So I guess that something in the middle is the way to go. Normally, on the projects I've worked in the past we use a single data set for each test class and I think that's a good common ground. In our specific case that we are working with openmrs data model, that almost never changes I think we have more headroom to make the data sets even more specific. In general I would prefer more specific data sets as it is easier to maintain and understand.

kaweesi: so, if am getting you well @edrisse, in this case u would prefer duplicating the same data if not defining new data when testing RoutineCalculation which runs both the other 2 calculations? i don't like huge datasets too and very much like more specific ones, the idea now is the balance hits at dependent calculations :-)

PR #48, opened 2019-02-06
MEPTS-87: Refactored 4month retention indicator

nthfloor:
refactored SQL queries to shift the data manipulation to the dataset and
replaced trasnferedin custom query with generic query aleady used for
trasnfered out. Then udpated all the parameters to pass the start adn
end dates down the line.


nthfloor:
@edrisse @ningosi please have a look at make any necessary adjustments

Something is off with how the dataset is interpreting the parameters, please review and see if you can spot the difference. 

The idea here is to use set the endDate and startDate(according to the user story) at the dataset level and use it everywhere else. Some edits had to be made to use pass the startDate through everywhere. I think I got everything for the _4month report_. If this change checks out, please do the same for the _2month report_.

PR #50, opened 2019-02-19
MEPTS-127 txPvls Criteria 2 bug fix

ynurmahomed:


api/src/main/resources/moduleApplicationContext.xml
```diff
@@ -55,4 +55,8 @@
             </list>
         </property>
     </bean>
+
+  <bean name="patientCalculationService"
```
kaweesi: this doesn't seem right, i am assuming you added this just for the sake of the test. this should be causing bean duplication with integration since calculation module adds the same

api/src/test/java/org/openmrs/module/eptsreports/reporting/calculation/pvls/RoutineCalculationTest.java
```diff
@@ -0,0 +1,312 @@
...
+import org.openmrs.calculation.result.ObsResult;
+import org.openmrs.calculation.result.SimpleResult;
+import org.openmrs.module.eptsreports.metadata.HivMetadata;
+import org.openmrs.module.eptsreports.reporting.calculation.BooleanResult;
+import org.openmrs.module.eptsreports.reporting.calculation.EptsCalculations;
+import org.openmrs.module.eptsreports.reporting.utils.EptsReportConstants.PatientsOnRoutineEnum;
+import org.openmrs.module.reporting.common.TimeQualifier;
+import org.openmrs.test.BaseContextMockTest;
+
+public class RoutineCalculationTest extends BaseContextMockTest {
```
kaweesi: can we move testing these just fixed criterias into https://github.com/esaude/openmrs-module-eptsreports/blob/master/api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/pvls/RoutineCalculationTest.java to be run as integrated tests rather. it will even make is easy to review them.

api/src/main/resources/moduleApplicationContext.xml
ynurmahomed: declared the bean here because it could not Autowired to `AbstractPatientCalculation`, but since we are going to stick with `Context.get` I'll revert it.

api/src/test/java/org/openmrs/module/eptsreports/reporting/calculation/pvls/RoutineCalculationTest.java
ynurmahomed: I don't know. The task says unit test. How will it make it easy to review?

nthfloor: Lets not replace the unit tests, but yes please extend the Routine integration tests by adding scenarios to cater for criteria2 & 3

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/pvls/RoutineCalculationTest.java
```diff
@@ -24,7 +26,7 @@ public PatientCalculation getCalculation() {
 
   @Override
   public Collection<Integer> getCohort() {
-    return Arrays.asList(new Integer[] {2, 6, 7, 8, 999, 432});
+    return Arrays.asList(new Integer[] {13, 6, 7, 8, 999, 432});
```
ynurmahomed: Had to create a new scenario because after the fix criteria2 matches for patient 2 and this scenario is shared with other tests.

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/calculation/pvls/RoutineCalculationTest.java
```diff
@@ -80,4 +82,32 @@ public void calculateRoutineCriteria1WhenBreastFeedingShouldMatchBetween3To6Mont
         getResult().toString(),
         service.evaluate(getCohort(), getCalculation(), params, getEvaluationContext()).toString());
   }
+
+  @Test
+  public void
+      calculateRoutineCriteria2WhenAdultsAndChildrenShouldMatchBetween12To15MonthsAndVlOfLessThan1000Copies() {
+    List<Integer> cohort = Arrays.asList(10);
```
ynurmahomed: Not sure how I am supposed to integrate this scenario in the result map.

ynurmahomed:
see #58

PR #51, opened 2019-02-20
MEPTS-61 -  Tech Debt: add unit tests for calculation utils

kaweesi:


api/src/test/java/org/openmrs/module/eptsreports/reporting/unit/utils/EptsCalculationUtilsTest.java
```diff
@@ -0,0 +1,378 @@
...
+  @Mock private PersonDataService personDataService;
+
+  @Mock private ServiceContext serviceContext;
+
+  private PatientCalculationContext calculationContext;
+
+  @SuppressWarnings("deprecation")
+  @Before
+  public void init() {
+    contextMockHelper.setService(PersonDataService.class, personDataService);
```
ynurmahomed: Forgot about this, looks like `ContextMockHelper` is deprecated. We should remove `Context.getService` calls from `EptsCalculationUtils`.

kaweesi: sure, i hated using it for the same cause, but it looks like the move is tending to use of `BaseModuleContextSensitiveTest` instead `BaseContextMockTest` whenever testing something that tends to do `Context.*` requests.

api/src/test/java/org/openmrs/module/eptsreports/reporting/unit/utils/EptsCalculationUtilsTest.java
```diff
@@ -0,0 +1,378 @@
...
+  @Test
+  public void ensureEmptyListResultsShouldReplaceNullsWithEmptyList() {
+    CalculationResultMap map = new CalculationResultMap();
+    ListResult list = new ListResult();
+    map.put(1, list);
+    map.put(2, null);
+    map.put(3, list);
+    Assert.assertNull(map.get(2));
+    CalculationResultMap replacedMap =
+        EptsCalculationUtils.ensureEmptyListResults(map, Arrays.asList(1, 2, 3));
```
nthfloor: Remember to add some test for those edge cases (pararmeters are null etc...)

kaweesi: of-couse this won't test NPE, closely examine each entry in the map since each represents a scenario

kaweesi:
- thanks for catching that, i must have just forgotten that unit test
- addMonths() looks exactly what it is which probably could explain why its author never added javadocs to it, we can probably.
- unit tests only mock external dependencies, there's no way of unit testing those internal non accessible functions outside the util except to unit test their functions evaluating them. the evaluateWithReporting unit tests are well setup to cover those internal methods.
- each of those cases test externally accessible methods, leaving them non unit tested means if someone ever modifies them from their simple invocation we see in them, we won't have any tests to help us be sure the edit is valid.
- i get no compile time errors, if so travis would still be catching those, please share logs if am missing something.

nthfloor:
* I meant to say compile-time _warnings_ instead of errors - have a look at your log when the module compiles.
* Remove encounterResultForPatient and obsResultForPatient since they just pipe though to resultForPatient from EptscalculationUtils file itself. Not just the tests. These functions are not adding any value.
* For  (toCalculationResult, ensureReportingContext), you're right, I keep thinking in JavaScript! :man_facepalming: In JS when you import a file you get access to all its functions so you can test them. Anyways never mind about this comment for these functions.

kaweesi:
- cool, i will take a look sooner
- if we are removing them from the util, then that makes sense.

api/src/test/java/org/openmrs/module/eptsreports/reporting/unit/utils/EptsCalculationUtilsTest.java
ynurmahomed: I don't think it will be necessary to use `BaseModuleContextSensitiveTest`. I can see only 2 calls to `Context.getSevice` in `EptsCalculationUtils` which are already mocked/stubbed in these tests.

I think we only need to define the services as fields and some king of injection like they describe here: 
https://github.com/openmrs/openmrs-core/blob/4a0feb8da351088f25fdc4e6d324a1f277aa3410/api/src/test/java/org/openmrs/api/context/ContextMockHelper.java#L42-L55

kaweesi: just switched to run with powermock so as to statically mock Context.

api/src/main/java/org/openmrs/module/eptsreports/reporting/utils/EptsCalculationUtils.java
```diff
@@ -24,7 +24,6 @@
 import org.joda.time.DateTime;
```
nthfloor: @kaweesi please move these inline context.get...calls to the top as global variables(`evaluateWithReporting(...)` needs to be updated). Easier to see what the dependencies are for this class.

PR #52, opened 2019-02-20
Disabled openmrs formatter

nthfloor:
Openmrs default formatter is using an older plugin which is has not been
maintained. We have replaced with a newer plugin and are disabling the
old one so the code does not get reformatted twice each time we compile the module.


kaweesi:
looks nice, 👍

PR #55, opened 2019-02-27
MEPTS-151: Implement Tx_TB indicator

kaweesi:


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXTBQueries.java
```diff
@@ -0,0 +1,298 @@
...
+      Integer ARVPlanConceptId,
+      Integer startDrugsConceptId,
+      Integer historicalDrugsStartDateConceptId,
+      Integer ARTProgramId,
+      Integer pharmacyEncounterTypeId,
+      Integer artAdultFollowupEncounterTypeId,
+      Integer artPedFollowupEncounterTypeId) {
+    return "SELECT patient_id FROM (SELECT patient_id, Min(data_inicio) data_inicio "
+        + "FROM (SELECT p.patient_id, Min(e.encounter_datetime) data_inicio FROM patient p "
+        + "INNER JOIN encounter e ON p.patient_id = e.patient_id "
```
edrisse: @kaweesi we should be using `String.format()` on these queries to make them more readable and easier to copy to an SQL client when needed during investigations, development, etc. It's something that has been very useful in the past.

api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/SetupMERSemiAnnualReport.java
```diff
@@ -0,0 +1,95 @@
...
+  @Override
+  public String getDescription() {
+    return "MERSemiAnnual - PEPFAR MER 2.1";
+  }
+
+  @Override
+  public ReportDefinition constructReportDefinition() {
+    ReportDefinition reportDefinition = super.constructReportDefinition();
+
+    reportDefinition.setBaseCohortDefinition(
```
edrisse: @kaweesi Shouldn't we be using `genericCohortQueries.getBaseCohort()` as base cohort?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXTBQueries.java
kaweesi: thanks a lot @edrisse for your review, if i was making a choice on readability, then not using String.format() would be more readable to me. maybe easily copying the query but still u would deal with '+' ... etc

api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/SetupMERSemiAnnualReport.java
kaweesi: are these 2 doing the same, probably using that base cohort would have broken my former matching figures with existing report, glad we are on changes

edrisse: That is the base cohort we are using on all reports. It comes from the SQL reports from FGH. I think you should use it on this code.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXTBQueries.java
edrisse: @kaweesi I think you should change the code. We already have refactored the older reports to use `String.format()` and merging this code the way it is would be a step backwards.

kaweesi: besides those 2 reasons, are there other reasons that were considered for the refactoring?

api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/SetupMERSemiAnnualReport.java
kaweesi: because of the changes we are already working on, i changed this, i kind of think this change may have affected the old report had it been a match with old numbers that we were considering, am however not sure if all reports had the same base cohort before, if so then probably that wouldn't have affected the matchings

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXTBQueries.java
edrisse: Yes, we also started using the dimensions in a proper way (we were using composition queries) and moved the base cohort from the query definitions to the report definition. Also we now tend to use calculations when possible.

PR #56, opened 2019-03-04
Correct the breastfeeding and pregnant composition cohort definitions

ningosi:
@ynurmahomed this is the changes that I thought were mapped wrongly that need to be corrected. If you get time please review those


ningosi:
Closing so that I can merge those into `MEPTS-127` branch which after review and verifications, will be merged into `develop`

PR #64, opened 2019-03-22
Mepts 153 implement tx ml indicator

ningosi:
TXML indicator development


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TxMlQueries.java
```diff
@@ -0,0 +1,64 @@
...
+  public static String getPatientsWhoMissedAppointment(
+      int min,
+      int max,
+      int returnVisitDateForDrugsConcept,
+      int returnVisitDate,
+      int pharmacyEncounterType,
+      int adultoSequimento,
+      int arvPediatriaSeguimento) {
+
+    return "SELECT patient_id FROM (SELECT p.patient_id,MAX(encounter_datetime) encounter_datetime FROM patient p "
```
edrisse: @ningosi i recommend we use `String.format()` here to make this queries more readable

api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/PepfarMerTxMlNumeratorReport.java
```diff
@@ -0,0 +1,86 @@
...
+              reportDefinition,
+              "TPEPFAR_MER_23_SEMIANNUAL_TX_ML.xls",
+              "PEPFAR MER 2.3 SEMIANNUAL TX_TB",
+              getExcelDesignUuid(),
+              null);
+      Properties props = new Properties();
+      props.put("sortWeight", "5000");
+      reportDesign.setProperties(props);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
```
edrisse: We should use `ReportException` here

PR #69, opened 2019-03-28
Fix error when running MER 2.3 report

edrisse:


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/dimensions/EptsCommonDimension.java
```diff
@@ -74,12 +74,12 @@ public CohortDefinitionDimension age(AgeDimensionCohortInterface ageDimensionCoh
     dim.addCohortDefinition(
         "5-9", ageDimensionCohort.createXtoYAgeCohort("patients with age between 5 and 9", 5, 9));
     dim.addCohortDefinition(
-        "<15", ageDimensionCohort.createXtoYAgeCohort("patients with age below 15", null, 15));
+        "<15", ageDimensionCohort.createXtoYAgeCohort("patients with age below 15", 200, 15));
```
kaweesi: this is meant to be 1 to 14

kaweesi: i wonder why it doesn’t fail for the some other reports, can we make the function rather accept and treat those nulls are values of 0 and 120 respectively

nthfloor: yeah we have logic to check if max & min ages are null, but seems the underlying age cohort does not work well if we don't specify explicit age boundaries.

kaweesi: just thinking making org.openmrs.module.eptsreports.reporting.library.cohorts.AgeCohortQueries.createXtoYAgeCohort and org.openmrs.module.eptsreports.reporting.library.cohorts.GenericCohortQueries.getAgeOnArtStartDate null accepting so as min and max placeholders can be set in there is better to save us future pain

edrisse: We just make sure the XML file on `serialized_object` table does not end up with `<null />` tags like the picture

xml-file

PR #70, opened 2019-03-28
Fgh live test utils

kaweesi:
- just storing these here for the future


ynurmahomed:
Could this be turned into a way to compare results and see if they match?

Maybe if we could load FGHs report definition, evaluate, get indicator results and compare with our report definitions...

I'm thinking about automating this part of the process

kaweesi:
this is built for that, evaluate definitions against real values, these values being the right counts maybe from FGH or ICAP or any other db, we can probably extend the idea so as to read the values from report template and populate our values into it, the tests would match against the right values and also auto insert for us our values to easy the time spent in moving values from the executed report into the online results matching template. very good suggestion @ynurmahomed

PR #73, opened 2019-03-29
MEPTS-157: Review & test TxNew and TxCurr

ynurmahomed:
This will enable us to consider patients wrongly registered with age > 200.
Use current year to be safe, bigger values might cause problems
with AgeCohortDefinition.

This solution needs to be revised.


ynurmahomed:
I have also updated the template for TX_CURR report

nthfloor:
Lets leave this for now - the scenario of patients older than 200 is an anomaly which should not happen in real life. But we need a more permanent solution for this which can form a tech debt ticket for next week.

nthfloor:
@ynurmahomed what did you update for TxCurr? Can you please commit that change to [this PR](https://github.com/esaude/openmrs-module-eptsreports/pull/74) since we are not going to merge this change for txNew.

ynurmahomed:
@nthfloor there were no changes for TX_CURR

PR #74, opened 2019-03-29
Pvls semiannual report fixes

esaude-ops:
* PVLS routine bug fix
* Update name of MER reports
* Fix tests


api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/pvls/RoutineCalculation.java
```diff
@@ -275,13 +275,14 @@ private boolean isOnRoutineCriteria1(
...
+            EptsCalculationUtils.monthsSinceIncludingDaysDiff(vlDate, artInitiationDate);
         if (criteria.equals(PatientsOnRoutineEnum.ADULTCHILDREN)
-            && (EptsCalculationUtils.monthsSince(artInitiationDate, vlDate) > 6
-                && EptsCalculationUtils.monthsSince(artInitiationDate, vlDate) <= 9)) {
+            && (monthsSince > 6 && monthsSince <= 9)) {
           return true;
         } else if (criteria.equals(PatientsOnRoutineEnum.BREASTFEEDINGPREGNANT)
-            && EptsCalculationUtils.monthsSince(vlDate, artInitiationDate) > 3
-            && EptsCalculationUtils.monthsSince(vlDate, artInitiationDate) <= 6) {
+            && monthsSince > 3
```
ningosi: Still we NOT including exactly 3 months, by doing  monthsSince >= 3? I think FGH does that, but Our specs exclude the exact match, otherwise all looks good

ningosi:
These fixes are tested in the **PEPFAR MER 2.3 Quarterly** right? The titile of this PR says **semiannual report** is that the same thing?

api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/pvls/RoutineCalculation.java
nthfloor: I am confirming with Pinki :+1:

nthfloor:
Sorry yes this branch has become a bit of a catch all for a few changes - we also had to update the name of the semiannual report.

But the PVLS change is specific to the quarterly report

ningosi:
On test, I have seen sample values for monthsSince flaot variable as 7.633333, 8.433333, 6.7, 6.9, 7.0333333, any danger to round off to say 2 decimal places? But still I see no harm in leaving them that way. Great job

ningosi:
@nthfloor all looks good to me, built locally and ran successfully, I will proceed to other investigations

api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/pvls/RoutineCalculation.java
nthfloor: Pinki says > 6 is fine, we can leave it.

PR #79, opened 2019-04-03
Bug fixs for TxTB numerator

nthfloor:
Bug1: in query for patients previously on TB in 6 month prior to
startDate, we were checking <=StartDate

Bug2: for same query we only needed to check for encounter types 6,9.
Encoutner types 20&26  nolonger being used

Bug3: perviousOnART we needed to check for <startDate not <= We updated
the parameter for this one.


nthfloor:
Needs to be regression tested across all dbs to be sure we match for all.

PR #82, opened 2019-04-11
MEPTS-84: unit test calculationcohort

ynurmahomed:


api/src/main/java/org/openmrs/module/eptsreports/reporting/cohort/evaluator/CalculationCohortDefinitionEvaluator.java
```diff
@@ -70,7 +72,7 @@ protected CalculationResultMap doCalculation(
       if (onDate == null) {
         onDate = (Date) context.getParameterValue("endDate");
         if (onDate == null) {
-          onDate = new Date();
+          onDate = clock.getCurrentDate();
```
kaweesi: any future plans for which this clock is gonna be used for?

ynurmahomed: No, I just added it to cover all scenarios. Current date will not be used anymore, we are sticking with `onOrBefore` ([MEPTS-171](https://jembiprojects.jira.com/browse/MEPTS-171)).

kaweesi: how about if we then get rid of that entire Clock resource!

ynurmahomed: Sure, I will have to remove the unit test also

ynurmahomed: @kaweesi done

PR #89, opened 2019-04-17
MEPTS-174: We exclude pregnant when checking breastfeeding

edrisse:


nthfloor:
@edrisse since we need this test for the patch release, can you please make this change off the master branch. We don't want to include those other changes that are in develop just yet.

edrisse:
Replaced by https://github.com/esaude/openmrs-module-eptsreports/pull/90 to be merged on master

PR #93, opened 2019-04-18
MEPTS-182: Fix TxCurr

edrisse:
value


nthfloor:
I have confirmed with Pinki that we can go ahead with committing this change, but it will need to be tested across all dbs to ensure the change does not affect existing scenarios. Pinki also confirmed this will not be released as a patch but will instead be combined with the June release. In which case please merge this PR into the develop branch.

Thanks,

PR #95, opened 2019-04-22
MEPTS-171 Refactor CalculationCohortDefinition

ynurmahomed:
CalculationCohortDefinition

 * onDate parameter changed to onOrBefore

 * added onOrAfter parameter

CalculationCohortDefinitionEvaluator

 * removed usages of date, endDate and currentDate. using only onOrBefore

 * Using cache to store onOrBefore instead of property now from CalculationContext


kaweesi:
+1 for the last commit

PR #101, opened 2019-05-17
alteracao da mensagem do ralatorio

sholiik:


nthfloor:
@sholiik just wanted to confirm if this was an intentional change or something from the training workshop?

sholiik:
Hi Nathan

It was intentional. I am trying to install openmrs on windows 10.

I was trying to use the openmrs.war that we use  on the training workshop because I’m facing problems during de deploy of openmrs.war version 1.11.5

Could you help me?

Regards.

sholiik

nthfloor:
Ok, this change is not going to help with solving that kind of setup issue. 

We can certainly assist with any setup issues you're having. But please, mainly for record purposes, post your issue on the Jembi helpdesk portal `https://helpdeskmoz.jembi.org`. We can then certainly help from there.

PR #103, opened 2019-05-31
Mepts 179 tx curr differences bux fix base cohort

ningosi:
Fixed the base cohort to meet the CCS databases


psmatsinhe:
Even being still in review we need to merge as there's a need to release a patch with this changes

PR #105, opened 2019-06-03
Mepts 195

ningosi:
Removing the Female option in the pregnancy and breastfeeding
Adjusting routine to match new updates


psmatsinhe:
Even being in review we need to merge as we releasing a patch for partners

PR #110, opened 2019-06-21
Mepts 213 pregnant breast feeding queries

KelvinWanyama:
All the pregnant and breastfeeding queries updated by removing the 'female' filter.


ningosi:
Merging these changes into MEPTS-179 so that we can test those along the base cohort change and routine changes

PR #115, opened 2019-07-25
PowerMock is not compatible with Java 11

kaweesi:
- https://issues.jenkins-ci.org/browse/JENKINS-55233


kaweesi:
wait, there could be license issues with oracle java, its commercial i hear

PR #119, opened 2019-09-03
Mepts 241 imer report fixes

KelvinWanyama:
- Removed obs with null date from the NextAndPrevDates Calculation class.

### DIFF
--- a/api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/dsd/NextAndPrevDatesCalculation.java
+++ b/api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/dsd/NextAndPrevDatesCalculation.java
@@ -63,7 +63,6 @@ public class NextAndPrevDatesCalculation extends AbstractPatientCalculation {
         Obs obs = iterator.next();
         if (obs.getVoided()
             || obs.getValueDate() == null
-            || obs.getValueDatetime() == null
             || obs.getEncounter() == null
             || (obs.getEncounter() != null
                 && (obs.getEncounter().getEncounterId() != lastEncounter.getEncounterId()))) {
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/dsd/NextAndPrevDatesCalculation.java
```diff
@@ -62,6 +62,8 @@ public CalculationResultMap evaluate(
       while (iterator.hasNext()) {
         Obs obs = iterator.next();
         if (obs.getVoided()
+            || obs.getValueDate() == null
```
nthfloor: getValueDate & getValueDateTime return the same object - do we need both?

KelvinWanyama: GetValueDateTime has the time component while getValueDate has only the date.

nthfloor: They return the same object, see here: https://github.com/openmrs/openmrs-core/blob/06593ec985c2e2064e13d172159b04f4398ca1ae/api/src/main/java/org/openmrs/Obs.java#L804-L823

KelvinWanyama: You are right @nthfloor We only need one, either return both date and time components.

KelvinWanyama: I will go ahead and make the changes.

PR #130, opened 2019-11-04
MEPTS-282: RM Dev: Updates in Tx-NEW IM-ER Report

ynurmahomed:


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxNewCohortQueries.java
```diff
@@ -13,26 +13,28 @@
  */
 package org.openmrs.module.eptsreports.reporting.library.cohorts;
 
+import static org.openmrs.module.reporting.evaluation.parameter.Mapped.mapStraightThrough;
```
cfaife: there is `@Autowired private HivCohortQueries hivCohortQueries` that is not  being  used,so it should be removed at line  `52`

cfaife:
PR review and  well done  for the work @ynurmahomed

PR #131, opened 2019-11-05
Mepts 284 rm dev changes in tx curr for im er report

cfaife:


cfaife:
yes  reviewed it... well done for the work @ynurmahomed

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXCurrQueries.java
```diff
@@ -0,0 +1,309 @@
...
+      int reasonPatientNotFound,
+      int noConcept,
+      int patientIsDead) {
+    String query =
+        "select  p.patient_id"
+            + " from patient p "
+            + " inner join encounter e on e.patient_id=p.patient_id "
+            + " inner join obs o  on o.encounter_id=e.encounter_id "
+            + " where  p.voided=0  and e.voided=0 and o.voided=0 "
+            + " and e.encounter_type in (%s) and o.concept_id in (%s,%s) and o.value_coded in (%s,%s) "
```
nthfloor: So this is checking that the patient has any concept matched with any value_coded, but the specs are actually more specific than that:
```Patient not found (Concept ID 2003) = NO (Concept ID 1066)
Reason of Not Finding (Concept ID 2031) = Died (COncpet Id 1383)```
So we should actually be pairing the concept with the value coded value - to be match the specs closer

File: api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXCurrQueries.java [hunk 2]
```diff
@@ -0,0 +1,309 @@
...
+      int aRVPharmaciaEncounterType,
+      int masterCardDrugPickupEncounterType,
+      int artDatePickup) {
+    String query =
+        " select p.patient_id "
+            + " from patient p "
+            + " inner join encounter e on  e.patient_id = p.patient_id "
+            + " where  p.voided=0  and e.voided=0 "
+            + " and e.encounter_type in (%s,%s,%s)  "
+            + " and e.encounter_datetime = (select  max(encounter_datetime) from encounter where  patient_id= p.patient_id) "
```
nthfloor: So the specs say: `encounter_datetime> the most recent date` 

this query will pick the most recent encounter, but not necessarily the most recent encounter for those encounter types. Have you confirmed with Pinki if we need to be considering any/all encounters or just those linked to those 3 encounter types?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXCurrQueries.java
```diff
@@ -0,0 +1,309 @@
...
+            + " and e.location_id= :location group by p.patient_id "
+            + " union "
+            + " select p.patient_id "
+            + " from patient p "
+            + " inner join encounter e on e.patient_id=p.patient_id "
+            + " inner join obs o on o.encounter_id=e.encounter_id "
+            + " where e.encounter_type = %s and p.voided=0  and e.voided=0 and o.voided=0 "
+            + " and o.concept_id=%s  "
+            + " and o.value_datetime > (select  max(value_datetime) from obs where  encounter_id= e.encounter_id) "
```
nthfloor: Same comment here.

In fact here, I am pretty sure def needs to be limited to the concept id, otherwise we could be comparing the value_date for a different obs

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
```diff
@@ -120,54 +149,141 @@ public CohortDefinition getTxCurrCompositionCohort(String cohortName, boolean cu
...
+            "14",
+            EptsReportUtils.map(
+                getPatientWithoutScheduledDrugPickupDateMasterCardAmdArtPickup(),
+                "location=${location}"));
 
     String compositionString;
     if (currentSpec) {
       compositionString =
-          "(1 OR 2 OR 3 OR 4) AND (NOT (5 OR ((6 OR (NOT 11)) AND (NOT (7 OR 8))))) AND (11 OR 12)";
+          "(1 OR 2 OR 3 OR 4 OR 5) AND NOT ((6 OR 7 OR 8 OR 9 OR 10 OR 11) AND NOT 12) AND NOT (13 OR 14)";
```
nthfloor: I think we need change the brackets in the composition string here.

Pinki left a note that the order of hte exclusions is important, i.e. we need to exclude the patients who left ART before then excluding the LTFU patients.

So soemthing like: `(main query AND NOT patient who left) NOT LTFU...etc...`

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXCurrQueries.java
cfaife: I  think here  I missed the last  encounter  as the spec  says....

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXCurrQueries.java
nthfloor: @cfaife I see the max encounter date query is still not filtering on encounter type - have you confirmed with Pinki if this is ok?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXCurrQueries.java
nthfloor: @cfaife same here, you haven't changed this yet

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXCurrQueries.java
```diff
@@ -306,4 +322,37 @@ public static String getPatientWithoutScheduledDrugPickupDateMasterCardAmdArtPic
...
+      int aRVPharmaciaEncounterType,
+      int masterCardDrugPickupEncounterType,
+      int artDatePickup) {
+    String query =
+        " select p.patient_id "
+            + " from patient p "
+            + " inner join encounter e on  e.patient_id = p.patient_id "
+            + " where  p.voided=0  and e.voided=0 "
+            + " and e.encounter_type in (%s,%s,%s)  "
+            + " and e.encounter_datetime > (select  max(encounter_datetime) from encounter where  patient_id= p.patient_id ) "
```
nthfloor: @cfaife this inner query is not filtering on encounter type - is that intentional?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXCurrQueries.java
```diff
@@ -306,4 +322,37 @@ public static String getPatientWithoutScheduledDrugPickupDateMasterCardAmdArtPic
...
+            + " and e.location_id= :location group by p.patient_id "
+            + " union "
+            + " select p.patient_id "
+            + " from patient p "
+            + " inner join encounter e on e.patient_id=p.patient_id "
+            + " inner join obs o on o.encounter_id=e.encounter_id "
+            + " where e.encounter_type = %s and p.voided=0  and e.voided=0 and o.voided=0 "
+            + " and o.concept_id=%s  "
+            + " and o.value_datetime > (select  max(value_datetime) from obs where  encounter_id= e.encounter_id) "
```
nthfloor: @cfaife this inner query is not filtering on obs.concept_id - is that intentional?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXCurrQueries.java
cfaife: done

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TXCurrQueries.java
cfaife: done

PR #149, opened 2019-11-26
MEPTS-292: RM: DSD Indicators Update - N3

ynurmahomed:


omod/src/main/resources/config.xml
```diff
@@ -1553,6 +1553,7 @@
 		</description>
 	</globalProperty>
 
+>>>>>>> develop
```
nthfloor: Seems this came from a merge resolution...?

nthfloor (changes requested):
@ynurmahomed seems you still need to remove the breastfeeding & pregnancy disaggs - pls remove that.

nthfloor (approved):

PR #157, opened 2019-12-04
Mepts 301

ningosi:
The new indicator TX-RTT to be included in PEPFAR MER 2.3 QUARTERLY REPORT.
Attached the USer Story and updated output template


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxRttCohortQueries.java
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * For each occurred consultation (Ficha de Seguimento Adulto, Pediatria or Ficha Clinica) of
+   * Encounter Type Ids 6 or 9 during the reporting period, the occurred encounter date minus the
+   * previous scheduled consultation date (Concept ID 1410 value_datetime) from the previous
+   * consultation of Encounter Type Ids 6 or 9, is greater than 28 days
+   *
+   * @return CohortDefinition
+   */
+  public CohortDefinition getAllPatientsWhoDelayedMoreThan28Days() {
```
cfaife: I think this method  should be private as it is not being using  on somewhere else

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxRttCohortQueries.java
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * * For each occurred drug pick ups of Encounter Types Ids 18 or 52 during the reporting period,
+   * the occurred encounter date minus the previous scheduled drug pick up date (as the most recent
+   * between previous encounter type 18 scheduled date (Concept ID 1410 value_datetime) and the
+   * previous encounter type 52 pick up date (Concept ID 23866 value_datetime +30 days)), is greater
+   * than 28 days
+   *
+   * @return CohortDefinition
+   */
+  public CohortDefinition getAllPatientsWhoDelayedMoreThan30Days() {
```
cfaife: the same  to this one , should be private

PR #159, opened 2019-12-05
MEPTS-298: RM: Updates in Tx-Curr (PEPFAR MER Quarterly Report) - Dec Release

ynurmahomed:


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
```diff
@@ -723,4 +728,171 @@ public CohortDefinition getPatientWhoAfterMostRecentDateHaveDrusPickupOrConsulta
...
+
+  private CohortDefinition getPatientsWithNextPickup83to173Days() {
+    return getPatientsWithNextPickupBetweenDaysAfterLastPharmacyEncounter(83, 173);
+  }
+
+  private CohortDefinition getPatientsWithNextPickupLessThan83days() {
+    return getPatientsWithNextPickupBetweenDaysAfterLastPharmacyEncounter(null, 83);
+  }
+
+  private CohortDefinition getPatientsWithMonthlyTypeOfDispensation() {
```
ningosi, line 807: Sometimes back we realized the reporting ORM was returning different values. Can you verify the results of this definition returns the same values as the SQL version? I know here the obs_datetime is used when filtering the last record, but when you looking for the last encounter, you will end up applying the encounter datetime, assuming the encounter date and obs date are similar for everyone, this might NOT have effect, but what if those are different? are we applying encounter date or obs date when filtering the last entry? Please ask

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
```diff
@@ -723,4 +728,171 @@ public CohortDefinition getPatientWhoAfterMostRecentDateHaveDrusPickupOrConsulta
...
+    cd.addParameter(new Parameter("locationList", "Location", Location.class));
+    cd.addEncounterType(hivMetadata.getAdultoSeguimentoEncounterType());
+    cd.setTimeModifier(BaseObsCohortDefinition.TimeModifier.LAST);
+    cd.setQuestion(hivMetadata.getTypeOfDispensationConcept());
+    cd.setOperator(SetComparator.IN);
+    cd.addValue(hivMetadata.getMonthlyConcept());
+    return cd;
+  }
+
+  private CohortDefinition getPatientsWithQuarterlyTypeOfDispensation() {
```
ningosi, line 822: Same here. Please ask which date are comparing against? Encounter date or obs datetime. CodedCohortDefinition will always look at the obs datetime check here https://github.com/openmrs/openmrs-module-reporting/blob/master/api/src/main/java/org/openmrs/module/reporting/cohort/definition/evaluator/BaseObsCohortDefinitionEvaluator.java

ningosi:
@ynurmahomed code looks fine. Please confirm on the date we are comparing against for the last outcome. After that clarity, we will run the results on develop and merge the code into develop

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TxCurrCohortQueries.java
ynurmahomed: For `FICHA CLINICA` form which uses `S.TARV: ADULTO SEGUIMENTO` encounter_type `encounter_datetime` is the same as `obs_datetime`.

PR #160, opened 2019-12-05
Mepts 299

ningosi:
PVLs changes, affect the numerator and the denominator
We have rewritten the routine criteria


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ViralLoadQueries.java
```diff
@@ -57,19 +57,108 @@ public static String getPatientsHavingViralLoadInLast12Months(
...
+            + "		INNER JOIN obs oo ON ee.encounter_id = oo.encounter_id"
+            + "	WHERE "
+            + "		ee.voided = 0 AND"
+            + "		oo.voided = 0 AND"
+            + "		ee.encounter_type = %d AND"
+            + "      oo.concept_id = %d AND"
+            + "      oo.value_coded IN(%d, %d, %d ,%d, %d, %d)"
+            + "		AND ee.location_id = :location"
+            + "		AND ee.encounter_datetime <= :endDate "
+            + ") e "
```
ynurmahomed: I think it is missing a `GROUP BY patient_id`.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ViralLoadQueries.java
```diff
@@ -57,19 +57,108 @@ public static String getPatientsHavingViralLoadInLast12Months(
...
+      int regimenFailureConceptId,
+      int suspectedImmuneFailureConceptId,
+      int repeatAfterBreastfeedingConceptId,
+      int clinicalSuspicionConceptId) {
+    String query =
+        "SELECT "
+            + "	e.patient_id "
+            + " FROM ("
+            + "	SELECT "
+            + "		ee.patient_id, MAX(DATE(ee.encounter_datetime)) AS viral_load_date"
```
ynurmahomed: I think `DATE()` is not necessary here.

api/src/test/java/org/openmrs/module/eptsreports/reporting/unit/library/cohorts/LiveTest.java
```diff
@@ -0,0 +1,67 @@
...
+  @Ignore
+  @Test
+  public void testMethod() throws EvaluationException {
+    CohortDefinition cd = hivCohortQueries.getPatientsWithViralLoadResults();
+    EvaluatedCohort evaluated = evaluateCohortDefinition(cd);
+    Set<Integer> patients = evaluated.getMemberIds();
+    for (Integer i : patients) {
+      System.out.println(i);
+    }
+    System.out.println("size:=> " + patients.size());
```
ynurmahomed: I think better not to commit live tests if they don't actually test anything.

omod/src/main/resources/config.xml
```diff
@@ -1351,7 +1351,16 @@
...
 		<description>
-			UUID for the Levantamento de ARV Master Card encounter type
+            UUID for the Levantamento de ARV Master Card encounter type
+		</description>
+	</globalProperty>
+
+	<!-- 51 -->
+	<globalProperty>
+		<property>eptsreports.fsrEncounterType</property>
+		<defaultValue>e2790f68-1d5f-11e0-b929-000c29ad1d07</defaultValue>
```
ynurmahomed, line 1361: What is the reason for mapping to same uuid as `MISAU: LABORATORIO`?

ynurmahomed (changes requested):
Looks good. There are some minor issues.

Also please have a look at breastfeeding and pregnant queries. 

According to the specs they should be the ones used in `TX_NEW` (`getPatientsPregnantEnrolledOnART` and `getTxNewBreastfeedingComposition`).

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ViralLoadQueries.java
ningosi: Adding group by clause here had a negative impact on the time it took to run the report. It was taking too long. If you have a second please try that code section directily on the DB with that clause on and without. I will include it though

omod/src/main/resources/config.xml
ningosi: I used that as a place holder since NOT all implementations have that encounter type present

api/src/test/java/org/openmrs/module/eptsreports/reporting/unit/library/cohorts/LiveTest.java
ningosi: I will remove this section.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ViralLoadQueries.java
ningosi: That removed

ningosi:
@ynurmahomed from the discussion I had with Pinki, we were to maintain the pregnant and breastfeeding the way they were per the last release. I have posted in slack for her views regarding those. If she says we change, I will change immediately and push for your review.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ViralLoadQueries.java
ynurmahomed: But without the `GROUP BY` the query is not correct.

PR #165, opened 2019-12-11
MEPTS-289: Updates in DSD indicators N2

ynurmahomed:


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/datasets/EriDSDDataset.java
```diff
@@ -383,27 +383,6 @@ public DataSetDefinition constructEriDSDDataset() {
...
-            mappings),
-        "");
-    dsd.addColumn(
-        "N3UPB",
-        "N3: Pregnant: includes breastfeeding patients",
-        EptsReportUtils.map(
-            eptsGeneralIndicator.getIndicator(
-                "N3UPB",
-                EptsReportUtils.map(
-                    eriDSDCohortQueries.getPatientsWhoArePregnantAndBreastfeedingN3(), mappings)),
```
nthfloor, line 404: @ynurmahomed I think we can remove the cohort `getPatientsWhoArePregnantAndBreastfeedingN3` since it isn't being used anywhere else, same comment for the breastfeeding cohort.

ynurmahomed: @nthfloor removed.

PR #168, opened 2019-12-17
MEPTS-194 Update  IM-ER4  with new LTFU, transfer out, dead and stop treatment disags.

Ekirapapaul:


Ekirapapaul:
Closing this PR after discussions with Pinki, we don't have to depend on txml queries for transfer , dead and stopped treatment diags.

PR #170, opened 2019-12-17
MEPTS-194 Updated ltfu disag on IMER-4 to exclude transfers, dead and…

Ekirapapaul:
… stopped treatment


moshonk (approved):
This looks good. Merging the PR

PR #173, opened 2019-12-19
Mepts 309 rm additional disaggregation tx ml

ItsMurumba:


api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
```diff
@@ -949,6 +949,14 @@ public Concept getAbsenceOfHealthProviderInHealthUnitConcept() {
     return getConcept(uuid);
   }
 
+  // concept_id = 1410
```
cfaife: Hi @KelvinWanyama this  is in https://github.com/esaude/openmrs-module-eptsreports/blob/bcc18f88ea5ef727c09022a1ad7fd3616564e4ef/api/src/main/java/org/openmrs/module/eptsreports/metadata/CommonMetadata.java#L104  so we do not need to add it in `HivMetadata.class`

cfaife (changes requested):

api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/txml/StartedArtOnLastClinicalContactCalculation.java
```diff
@@ -0,0 +1,126 @@
...
+   * @param context
+   * @return
+   */
+  private CalculationResultMap getLastClinicalContactMap(
+      Collection<Integer> cohort, PatientCalculationContext context) {
+    SqlPatientDataDefinition sqlPatientDataDefinition = new SqlPatientDataDefinition();
+    sqlPatientDataDefinition.addParameter(new Parameter("onOrBefore", "onOrBefore", Date.class));
+    sqlPatientDataDefinition.addParameter(new Parameter("location", "location", Location.class));
+
+    String query =
```
cfaife, line 83: why  you are  putting  pure query in  the Calculation, it would be better if  we use EptsCalculationUtils

cfaife (changes requested):

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TxMlQueries.java
```diff
@@ -33,6 +33,28 @@ public static String getPatientsWhoMissedAppointment(
         max);
   }
 
+  public static String getNonConsentedPatients(
```
cfaife, line 129: why you are joining with  Person?
it would be better if  join obs with encounters .... reducing inner joins to speedup the  query

api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/txml/StartedArtOnLastClinicalContactCalculation.java
```diff
@@ -0,0 +1,126 @@
...
+    CalculationResultMap artStartDates =
+        calculate(
+            Context.getRegisteredComponents(InitialArtStartDateCalculation.class).get(0),
+            cohort,
+            parameterValues,
+            context);
+    if (endDate == null) {
+      endDate = (Date) context.getFromCache("onOrBefore");
+    }
+    if (endDate != null) {
```
cfaife: this  `if`   is not needed

cfaife (changes requested):

api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/txml/StartedArtOnLastClinicalContactCalculation.java
ItsMurumba: We were embedding a query into a calculation. Got a recommendation from the team @nthfloor  and @ynurmahomed

api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/txml/StartedArtOnLastClinicalContactCalculation.java
ItsMurumba: I have removed the first if. I need to recheck and confirm endDate is not null.

api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
ItsMurumba: okay thanks.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TxMlQueries.java
ItsMurumba: Was reusing an existing query.

cfaife: yes but I think  we should   not join with person

ItsMurumba: I will have to ask Pinki for this user story then. Because it is a query from the old TxML

api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/txml/StartedArtOnLastClinicalContactCalculation.java
cfaife: we  have form  validation in the UI, do  you think we have to verify if endDate is null?
I'm quite sure  that endDate will always be not null.
image

cfaife (changes requested):

api/src/test/java/org/openmrs/module/eptsreports/reporting/unit/calculation/txml/StartedArtOnLastClinicalContactCalculationTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.openmrs.module.eptsreports.reporting.calculation.common.EPTSCalculationService;
+import org.openmrs.module.eptsreports.reporting.calculation.txml.StartedArtOnLastClinicalContactCalculation;
+import org.openmrs.module.eptsreports.reporting.helper.TestsHelper;
+import org.openmrs.module.eptsreports.reporting.unit.PowerMockBaseContextTest;
+import org.openmrs.module.eptsreports.reporting.utils.EptsCalculationUtils;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+
+@PrepareForTest(EptsCalculationUtils.class)
+public class StartedArtOnLastClinicalContactCalculationTest extends PowerMockBaseContextTest {
```
cfaife: this  class  has  no assertions

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TxMlQueries.java
```diff
@@ -144,4 +259,192 @@ public static String getRefusedOrStoppedTreatment(
...
+          int apoioReintegracaoParteAEncounterTypeId,
+          int apoioReintegracaoParteBEncounterTypeId) {
+
+    String query =
+        " SELECT pa.patient_id FROM patient pa "
+            + " WHERE pa.patient_id NOT IN ("
+            + "  SELECT pa.patient_id "
+            + "  FROM patient pa"
+            + "  INNER JOIN encounter e ON pa.patient_id=e.patient_id "
+            + "  INNER JOIN obs o ON pa.patient_id=o.person_id "
```
cfaife, line 364: why   join  obs  with person, couldn't it be encounter with  obs? in order to filter order obs by encounter_type.....something  like  ` INNER JOIN obs o ON o.encounter_id = e.encounter_id`

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TxMlQueries.java
cfaife: any feedback from Pinki here?

api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/txml/StartedArtOnLastClinicalContactCalculation.java
ItsMurumba: Done

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TxMlQueries.java
ItsMurumba: I think for now we can close this out and then Pinki will create a ticket to solve improve the query.

cfaife (changes requested):

PR #180, opened 2020-01-02
Mepts 315 investigate differences in pvls

ningosi:
This PR is meant to fix the differences in the numbers after comparing against FGH numbers


ningosi:
@psmatsinhe this request is waiting for Pinki to have a final test on the pvls numerator before we can merge into develop. Generally, all patients look to be matching from my end, unless new developments are reported

PR #183, opened 2020-01-09
MEPTS-321 added IF condition in Eri2Querie  to evaluate  the if  data…

cfaife:
…-inicion+33days is major than reportingEndDate


ningosi (commented):
Code looks good and well mapped. I will have it uploaded on the server and have the results

PR #184, opened 2020-01-16
MEPTS-326: Update LTFU disaggregation in IM ER4 only to consider 60 days instead of 30 days

ynurmahomed:


ningosi (approved):
Looks fine and report runs to completion

PR #188, opened 2020-01-23
MEPTS-334: Update Pregnant Queries/Functions

ningosi:
This ticket is ready for review, but results will available after the scripts are ready to be tested on different servers

### DIFF
--- a/api/src/main/java/org/openmrs/module/eptsreports/metadata/CommonMetadata.java
+++ b/api/src/main/java/org/openmrs/module/eptsreports/metadata/CommonMetadata.java
@@ -72,12 +72,6 @@ public class CommonMetadata extends Metadata {
         Context.getAdministrationService().getGlobalProperty("eptsreports.pregnantConceptUuid");
     return getConcept(uuid);
   }
-  // concept_id = 44
-  public Concept getGestationConcept() {
-    String uuid =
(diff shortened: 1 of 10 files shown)


api/src/main/java/org/openmrs/module/eptsreports/metadata/CommonMetadata.java
```diff
@@ -72,12 +73,6 @@ public Concept getPregnantConcept() {
     return getConcept(uuid);
   }
 
-  public Concept getGestationConcept() {
-    String uuid =
-        Context.getAdministrationService().getGlobalProperty("eptsreports.gestationConceptUuid");
```
cfaife, line 77: Hi @ningosi, do we no longer  need to use  44  concept?

ningosi: Yes, that concept is no longer required for the question having this concept_id 1982. So anywhere we had that question with an answer as 44, we replace with 1065. By doing that everywhere, I found that concept nolonger being used anywhere, that is why I ended up deleting it. But if there could be a usecase for it, it can be returned, but for me there were no uses needed for it

cfaife: ok thanks Nicholas

PR #189, opened 2020-01-24
MEPTS-294: rm dsd indicators new n5 fix mappings

ItsMurumba:
Fixed the mapping of the report on the excel.
Reverted the omitted logic for the ticket.


api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/SetupCombinedImErReport.java
```diff
@@ -5,10 +5,7 @@
 import java.util.List;
 import java.util.Properties;
 import org.openmrs.module.eptsreports.reporting.library.cohorts.GenericCohortQueries;
-import org.openmrs.module.eptsreports.reporting.library.datasets.Eri2MonthsDataset;
-import org.openmrs.module.eptsreports.reporting.library.datasets.Eri4MonthsDataset;
-import org.openmrs.module.eptsreports.reporting.library.datasets.TxCurrDataset;
-import org.openmrs.module.eptsreports.reporting.library.datasets.TxNewDataset;
+import org.openmrs.module.eptsreports.reporting.library.datasets.*;
```
ynurmahomed: @KelvinWanyama we should not be using wildcard imports, strange that the formatter allowed this.

ynurmahomed (changes requested):

api/src/main/java/org/openmrs/module/eptsreports/reporting/reports/SetupCombinedImErReport.java
ItsMurumba: Fixed this @ynurmahomed

PR #200, opened 2020-02-17
MEPTS-346: Change Request in Monthly Report (Resumo Mensal)

ynurmahomed:


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
```diff
@@ -187,7 +187,8 @@ public CohortDefinition getPatientsWhoInitiatedTarvAtThisFacilityDuringCurrentMo
    */
   public CohortDefinition
       getNumberOfPatientsTransferredInFromOtherHealthFacilitiesDuringCurrentMonthB2() {
-
+    // TODO maybe we should be re-using HivCohortQueries#getPatientsTransferredFromOtherHealthFacility
```
ningosi, line 190: Did you get any response from the BA

ynurmahomed: I got no response.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ResumoMensalQueries.java
```diff
@@ -63,24 +69,40 @@ public static String getPatientsTransferredFromAnotherHealthFacilityDuringTheCur
             + "         ON type.encounter_id = e.encounter_id "
             + "WHERE  p.voided = 0 "
             + "       AND e.voided = 0 "
-            + "       AND e.encounter_type = %d "
+            + "       AND e.encounter_type = ${masterCard} "
```
ningosi, line 72: What does this phrase means in the specs? **Select all patients with the earliest ART Start Date from the following criterias** Do you need to select all those dates per patient, then later take the first one(one that came first)/

ynurmahomed: I think it is, as implemented in `InitialArtStartDateCalculation`.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ResumoMensalQueries.java
```diff
@@ -63,24 +69,40 @@ public static String getPatientsTransferredFromAnotherHealthFacilityDuringTheCur
...
             + "       AND e.location_id = :location "
             + "       AND e.encounter_datetime BETWEEN :startDate AND :endDate "
             + "       AND transf.voided = 0 "
-            + "       AND transf.concept_id = %d "
-            + "       AND transf.value_coded = %d "
+            + "       AND transf.concept_id = ${transferFromOther} "
```
ningosi, line 76: The specs says **Exclude all patients transferred-in** based on those parameters, looks like we are getting sum of those patients in addition to the lower union. Are we NOt excluding those from the resultant query?

ynurmahomed: @ningosi I think you might be looking at specs for B1, this is B2.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ResumoMensalQueries.java
ningosi: in that class tarnslated to sql would be **UNION of all those subqueries(by getting patient_id and dateRequired, duplicates will be allowed here), then later have an outer query that  select patientId with MIN(dateRequired), here unique rows will be picked, then fhave an outer query that  pick a set of patientIDs, That is how I would translate that class to sql**

ningosi (approved):
The change matches what B2 needs. Thanks for clarification @ynurmahomed

PR #209, opened 2020-02-21
MEPTS-344: Changes in Monthly Report MISAU- A2

Edgencio:


api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
```diff
@@ -956,6 +956,14 @@ public Concept getBpostiveConcept() {
     return getConcept(uuid);
   }
 
+  // concept_id = 1369
+  public Concept getTransferredFromOtherFacilityConcept() {
```
ynurmahomed: There is already a method that returns this concept in, no need to re-implement:
https://github.com/esaude/openmrs-module-eptsreports/blob/019197fc89034447ec70eb041796bf0903757f1d/api/src/main/java/org/openmrs/module/eptsreports/metadata/CommonMetadata.java#L47-L52

omod/src/main/resources/config.xml
```diff
@@ -83,6 +83,14 @@
 		</description>
 	</globalProperty>
 
+	<globalProperty>
+		<property>@MODULE_ID@.preTarvConceptUuid</property>
```
ynurmahomed, line 87: I'm not sure if this global property is for `YES` or `PRE-TARV`. The uuid is for `YES` but it is already defined in the file.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
```diff
@@ -111,17 +111,87 @@ public CohortDefinition getPatientsWhoInitiatedPreTarvAtAfacilityDuringCurrentMo
     cd.addSearch(
         "A2I",
         map(sqlCohortDefinition, "startDate=${startDate},endDate=${endDate},location=${location}"));
-    cd.addSearch("A2III", map(getPatientsWithTransferFromOtherHF(), "locationList=${location}"));
     cd.addSearch(
         "A2II",
         map(
-            getPatientsWithFirstClinicalConsultationOnTheSameDateAsPreArtStartDate(),
-            "endDate=${endDate},location=${location}"));
+            getPatientsEnrolledInPreArtOrRegisteredInClinicalProcessOpening(),
```
ynurmahomed: From the spec:
> ◦ Select all patients registered in encounter “Master Card – Ficha Resumo” (encounter id 53) who have the following conditions: 
> ▪ “Pre-ART Start Date” (PT”: “Data do Inicio Pre-TARV”) (Concept ID 23808)
> ▪ enrolled in PRE-ART Program (program_id=1, from patient program table) with enrollement date (date_enrolled) OR
> ▪ registered in “Abertura de Processo Clínico” (encounter type id 5 or 7) with “Data de Registo” (encounter_datetime) 
> ▪ **with the most earliest date from above criterias in the reporting period (>=startDate and <=endDate)**

Implementation is missing the part highlighted in bold. 

You will probably need to do a union of `A2I` and `A2II` in a single query, check which of the criterias has the minimum date then check if this date falls between the reporting period.

You can use this part of a query as reference:
https://github.com/esaude/openmrs-module-eptsreports/blob/019197fc89034447ec70eb041796bf0903757f1d/api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TxMlQueries.java#L250-L295

This one looks for the most recent date between 3 criterias and adds 28 days to the date found.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
```diff
@@ -111,17 +111,87 @@ public CohortDefinition getPatientsWhoInitiatedPreTarvAtAfacilityDuringCurrentMo
...
+            getAllPatientsWithPreArtDateTransferredInFromWithBoundaries(),
+            "startDate=${startDate},endDate=${endDate},location=${location}"));
+    cd.addSearch(
+        "A2VI",
+        map(
+            getAllPatientsRegisteredAsTransferredInProgramEnrolmentWithBoundaries(),
+            "startDate=${startDate},endDate=${endDate},location=${location}"));
 
-    cd.setCompositionString("(A2I AND A2II) AND NOT A2III");
+    cd.setCompositionString("(A2I AND A2II) AND NOT (A2III AND A2IV AND A2V AND A2VI)");
```
ynurmahomed: `A2III`, `A2IV` and `A2V` will probably need to be combined to a single query. It does not say in the spec but all obs need to be in the same encounter (BA can confirm this) and using separate queries does not guarantee that.

It should look similar to this query:
https://github.com/esaude/openmrs-module-eptsreports/blob/45eeb243a29bd4e180b92bd463c7b3def619dc02/api/src/main/java/org/openmrs/module/eptsreports/reporting/cohort/evaluator/EptsTransferredInCohortDefinitionEvaluator.java#L40-L94

ynurmahomed (changes requested):
Very good first PR. There are issues, but those are mostly related to the specs.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ResumoMensalQueries.java
```diff
@@ -40,38 +40,43 @@ public static String getAllPatientsWithPreArtStartDateWithBoundaries(
...
-
+  /**
+   * All patients Enrolle in Pre-Art or Registered In Clinical Process Opening enddate
+   *
+   * @return String
+   */
+  public static String getAllPatientsRegisteredAsTransferredInProgramEnrolmentWithBoundaries(
+      int stateId, int programId) {
+    String query =
+        "SELECT p.patient_id FROM patient p INNER JOIN patient_program pp ON p.patient_id = pp.patient_id INNER JOIN patient_state ps ON ps.patient_program_id = pp.program_id WHERE p.voided=0 AND ps.state=%d AND pp.program_id = %d AND pp.location_id=:location pp.date_enrolled BETWEEN :startDate AND :endDate";
```
psmatsinhe: @Edgencio I see you doing may have some mistake here **### _

> ps.patient_program_id = pp.program_id

_**

psmatsinhe: @Edgencio can you please change this %d to the new approach

psmatsinhe (changes requested):

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ResumoMensalQueries.java
```diff
@@ -37,11 +37,132 @@ public static String getAllPatientsWithPreArtStartDateLessThanReportingStartDate
...
+            + "               e.encounter_datetime AS enrollment_date "
+            + "        FROM   patient p "
+            + "               INNER JOIN encounter e "
+            + "                       ON p.patient_id = e.patient_id "
+            + "               INNER JOIN obs o "
+            + "                       ON o.encounter_id = e.encounter_id "
+            + "        WHERE  p.voided = 0 "
+            + "               AND e.voided = 0 "
+            + "               AND o.voided = 0 "
+            + "               AND e.encounter_type = ${encounterType} "
```
ynurmahomed: Please use more descriptive names, the ideia is to be able to look at the query and know whats it is about without having to find out how it is mapped. 
Better to use `${masterCardEncounterType}`.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ResumoMensalQueries.java
```diff
@@ -37,11 +37,132 @@ public static String getAllPatientsWithPreArtStartDateLessThanReportingStartDate
...
+            + "        FROM   encounter enc "
+            + "               JOIN patient p "
+            + "                 ON p.patient_id = enc.patient_id "
+            + "        WHERE  p.voided = 0 "
+            + "               AND enc.encounter_type IN (${encounterType1},${encounterType2}) "
+            + "               AND enc.location_id =:location "
+            + "               AND enc.encounter_datetime BETWEEN"
+            + "                   :startDate AND :endDate "
+            + "        ORDER  BY enrollment_date ASC) results "
+            + "GROUP  BY results.patient_id) res  ";
```
ynurmahomed: Almost there. The algorithm is to look for the earliest date then check if it falls within the period. 
So, the date filter should **only** be in the first inner query.

First add an alias to the earliest date:
`Min(results.enrollment_date) enrollment_date`

Then filter by date:
`WHERE results.enrollment_date BETWEEN :startDate AND :endDate`

Don't forget to remove the filters from the inner most queries.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
ynurmahomed: `EptsTransferredInCohortDefinition` implements the program clause. 

So `getAllPatientsRegisteredAsTransferredInProgramEnrolmentWithBoundaries` is not necessary anymore, it should be deleted.

api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/library/cohorts/ResumoMensalCohortQueriesTest.java
```diff
@@ -58,4 +57,46 @@ protected void setParameters(
...
+    parameters.put(new Parameter("onOrAfter", "", Date.class), getStartDate());
+    parameters.put(new Parameter("onOrBefore", "", Date.class), getEndDate());
+    parameters.put(new Parameter("location", "", Location.class), getLocation());
+    EvaluatedCohort evaluatedCohort = evaluateCohortDefinition(cohort, parameters);
+    assertEquals(1, evaluatedCohort.getMemberIds().size());
+    assertTrue(evaluatedCohort.getMemberIds().contains(1010));
+  }
+
+  @Test
+   // TODO add some mappings on the dataset to properly test the whole query
```
ynurmahomed: Tests not implemented should be annotated with `@Ignore`.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ResumoMensalQueries.java
```diff
@@ -37,11 +37,132 @@ public static String getAllPatientsWithPreArtStartDateLessThanReportingStartDate
...
+    StringSubstitutor stringSubstitutor = new StringSubstitutor(map);
+    return stringSubstitutor.replace(query);
+  }
+
+  /**
+   * Number of patients transferred-in from another HFs during the current month
+   *
+   * @return String
+   */
+  public static String getPatientsTransferredFromAnotherHealthFacilityDuringTheCurrentMonth(
```
ynurmahomed: This method should be deleted because there are no usages.

ynurmahomed (changes requested):

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ResumoMensalQueries.java
Edgencio: Hey @ynurmahomed , i'm not the one who implemented that method... that's why i prefered to not mess with it! But im going to remove it now!

ynurmahomed: @Edgencio  no problem in removing, unused code should never be commited.

PR #211, opened 2020-02-24
Mepts 366 bug fix in data quality report

ningosi:
The data reports dates needed to be adjusted so that it excludes boundaries


psmatsinhe (approved):
Looks good to me

PR #221, opened 2020-02-25
Fix B5

ynurmahomed:


ningosi:
Some file conflicting here. Can you fix those?

PR #224, opened 2020-02-25
MEPTS-340: Adding updates to PEPFAR MER 2.4 Semi-Annual

rubailly:
Note: We are doing this together with Numerator (MEPTS-353) since the changes are very similar.
So please look at both MEPTS-340 and MEPTS-353 when reviewing this.


ynurmahomed:
@rubailly Has this change been implemented?
    

> 2. Compose indicator ART_list by 
> a. Selecting all patients who are ~~on ART by start of previous reporting period~~ by start of reporting period:

ynurmahomed:
@rubailly Clarified that the the change is only a change in wording on the document, does not affect actual code.

PR #228, opened 2020-02-27
MEPTS-362-RM: Changes in Monthly Report MISAU- B12

amosh101:


ningosi:
Looks like tests are failing on this PR, would you please fix them before I start to review?

ningosi:
Fix the conflicts, then I can review this work

amosh101:
> Fix the conflicts, then I can review this work

Hi Nicholas, done, please proceed.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
```diff
@@ -360,7 +413,7 @@ public CohortDefinition getPatientsWhoWereActiveByEndOfPreviousMonthB12() {
     CohortDefinition transferredOut = getPatientsTransferredOut();
     CohortDefinition suspended = getPatientsWhoSuspendedTreatment();
     CohortDefinition missedDrugPickup = getLastArvPickupDateCohort();
-    CohortDefinition died = getPatientsWhoDied();
+    CohortDefinition died = getPatientsWhoDied(false);
 
     String encounterWithCodedObsMappings = "onOrBefore=${startDate},locationList=${location}";
     String drugPickupMappings =
```
ningosi: Not used variable

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
```diff
@@ -360,7 +413,7 @@ public CohortDefinition getPatientsWhoWereActiveByEndOfPreviousMonthB12() {
     CohortDefinition transferredOut = getPatientsTransferredOut();
     CohortDefinition suspended = getPatientsWhoSuspendedTreatment();
     CohortDefinition missedDrugPickup = getLastArvPickupDateCohort();
```
ningosi: Not used variable

ningosi:
I tried looking through and came across some gaps as they are in specs:
1. B2A - The first section should have 3 conditions with 2 having value coded to question and the 3 one. Do all of them compare against obs_datetime? or only the last one. Are others based on encounter_datetime? Please confirm with Pinki. Also are those **UNOINS** OR **INTERSECTIONS**

ningosi:
And adding this to the patient state **AND ps.end_date is nul**l, means that you are picking the last one that is still active, are picking any of those or the last one? if we are picking the last one seems like even if we get one that is before the start date and end date is filled, then we will NOT be including them?

ningosi:
Other than having classes like **ResumoMensalLiveTest** which are ignored and doing nothing, can they be removed?

amosh101:
> Other than having classes like **ResumoMensalLiveTest** which are ignored and doing nothing, can they be removed?

Yes, fine by me, they were there already there and sitting unused.

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
amosh101: Removed

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
amosh101: Removed

ningosi:
B6A - the green highted text is saying **reporting start date** but from the code we are doing between startDate and endDate, NOT sure if those means the same thing

amosh101:
> B6A - the green highted text is saying **reporting start date** but from the code we are doing between startDate and endDate, NOT sure if those means the same thing

New B6 ticket still in testing.

amosh101:
> And adding this to the patient state **AND ps.end_date is nul**l, means that you are picking the last one that is still active, are picking any of those or the last one? if we are picking the last one seems like even if we get one that is before the start date and end date is filled, then we will NOT be including them?

with end date null we intend to indicate the patients must still be active in program wth transfered in state.

ningosi:
B6A - cd.addSearch("B6A", map(suspended, "onOrBefore=${startDate},location=${location}")); do  NOT match the mappings as are supplied in the sql cohort definition

amosh101:
> B6A - cd.addSearch("B6A", map(suspended, "onOrBefore=${startDate},location=${location}")); do NOT match the mappings as are supplied in the sql cohort definition

Understood Nicholas, i'll pay more attention next time

ningosi:
After few corrections, B12 indicator runs fine, will be considering to merge it into develop so that it fixes other errors reported in B6 parameter mappings

ningosi:
But this class **CodedObsOnFirstOrSecondEncounterCalculation** which is NOT releted to this code, need to be investigated. I will be waiting for values to output, then merge the code @amosh101 is there anything you think need to be added other that what we have discussed?

amosh101:
Thanks. This is it for now, we’ll have to see how it plays with the remaining subqueries that are being or will be updated.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #230, opened 2020-02-27
Mepts 361 c1

ningosi:
The C1 section of resumo mensal. C1: Number of patients who initiated Pre-TARV during the current month and was screened for TB


api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
```diff
@@ -1426,6 +1426,12 @@ public Concept getBPlusConcept() {
         Context.getAdministrationService().getGlobalProperty("eptsreports.bPlusConceptUuid"));
   }
 
+
+  //6275
+  public Concept getHivCareConcept() {
```
cfaife: this concept is repeated in this   file

Edgencio: There's already a concept method to fetch the HIV Care Concept it is called getPreTARVConcept(). To avoid duplications, maybe we remove the old one and and replace its occurrences with the new one (getHivCareConcep()t).

omod/src/main/resources/config.xml
```diff
@@ -1850,6 +1850,15 @@
 		</description>
 	</globalProperty>
 
+	<!--6275-->
+	<globalProperty>
+		<property>eptsreports.hivCareConceptUuid</property>
```
cfaife: this uuid is repeated

cfaife (changes requested):

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
```diff
@@ -598,6 +638,35 @@ private CohortDefinition getPatientsWithTransferFromOtherHF() {
     return cd;
   }
 
+  public CohortDefinition getAdditionalExclusionCriteriaForC1() {
+    CompositionCohortDefinition cd  = new CompositionCohortDefinition();
+    cd.setName("All patients to be excluded for the C1 definition");
+    cd.addParameter(new Parameter("onOrAfter", "onOrAfter", Date.class));
+    cd.addParameter(new Parameter("onOrBefore", "onOrBefore", Date.class));
+    cd.addParameter(new Parameter("locationList", "location", Location.class));
+    List<Concept> concepts = new ArrayList<>();
```
Edgencio: Looks like this **concepts** arrayList is an unused declaration, if so, maybe it should be removed

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
```diff
@@ -479,6 +495,30 @@ public CohortDefinition getPatientsWhoInitiatedPreTarvDuringCurrentMonthAndScree
...
+   * @return
+   */
+  public CohortDefinition getPatientsWhoInitiatedPreTarvDuringCurrentMonthAndScreenedTbC1() {
+    CompositionCohortDefinition cd = new CompositionCohortDefinition();
+    cd.setName("Patients who initiated Pre-TARV during the current month and was screened for TB C1");
+    cd.addParameter(new Parameter("startDate", "Start Date", Date.class));
+    cd.addParameter(new Parameter("endDate", "End Date", Date.class));
+    cd.addParameter(new Parameter("location", "Location", Location.class));
+
+    CohortDefinition c1 = getPatientsWhoInitiatedPreTarvAtAfacilityDuringCurrentMonthA2();
```
Edgencio: Here we also have an unused var **c1**, And also i believe that in the search where where you map the **Pop** key should be passed the *c1*. But let's maybe discuss this!

Edgencio (changes requested):

api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
ningosi: fixed

omod/src/main/resources/config.xml
ningosi: fixed

api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
ningosi: Maintained the old one. Just reused already existing

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
ningosi: removed

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
ningosi: Since **getPatientsWhoInitiatedPreTarvAtAfacilityDuringCurrentMonthC1** contains the values needed for c1, and I have used that directly, I will just remove the c1 and maintain the definition as it is

PR #231, opened 2020-02-27
Mepts c2

ningosi:
C2: Number of patients who initiated Pre-TARV during the current month and started TPI


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
```diff
@@ -598,6 +638,50 @@ private CohortDefinition getPatientsWithTransferFromOtherHF() {
...
+    List<Concept> concepts = new ArrayList<>();
+    cd.addSearch("transferBasedOnObsDate", map(getPatientsTransferBasedOnObsDate(), "onOrAfter=${onOrAfter},onOrBefore=${onOrBefore},locationList=${locationList}"));
+    cd.addSearch("getTypeOfPatientTransferredFrom", map(getTypeOfPatientTransferredFrom(), "onOrAfter=${onOrAfter},onOrBefore=${onOrBefore},locationList=${locationList}"));
+    cd.addSearch("inProgramState", map(genericCohortQueries.getPatientsBasedOnPatientStates(hivMetadata.getHIVCareProgram().getProgramId(),
+            hivMetadata.getPateintTransferedFromOtherFacilityWorkflowState().getProgramWorkflowStateId()), "startDate=${onOrAfter},endDate=${onOrBefore},location=${locationList}"));
+    cd.setCompositionString("transferBasedOnObsDate OR getTypeOfPatientTransferredFrom OR inProgramState");
+    return cd;
+  }
+
+  public CohortDefinition getAdditionalExclusionCriteriaForC2() {
```
Edgencio: Looks Like the **getAdditionalExclusionCriteriaForC2()** is similar to **getAdditionalExclusionCriteriaForC1()**, maybe you could create one function with a generic name and reuse it#

Edgencio (changes requested):

omod/src/main/resources/config.xml
```diff
@@ -1850,6 +1850,15 @@
 		</description>
 	</globalProperty>
 
+	<!--6275-->
+	<globalProperty>
+		<property>eptsreports.hivCareConceptUuid</property>
```
Edgencio: There is a duoplication of this concept... More info about this, on the change requests for the **MEPTS_C1** Pull Request.

Edgencio (changes requested):

omod/src/main/resources/config.xml
ningosi: If C1 is merged, all these will be sorted out

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
ningosi: The difference comes in the parameter manipulation, one needs one month be sutracted from startDate for both 2 searches and one just maintain. I would suggest those to stay that way for now

ningosi:
@cfaife this ticket relates to the one you just reviewed and merged. I will be fixing the conflicts and code refinement, then I have it be checked. The difference is so little compared to C1, then later we will finish with C3 before the entire ticket.

ningosi:
@cfaife when you get time check on this one. I need to finish all the C sections

PR #236, opened 2020-03-04
MEPTS-369: RM-Change Request in Resumo Mensal - B7

Edgencio:


rubailly (commented):
@Edgencio Where is the check for **patients who Abandoned the ART during the current month** ?
According to the specs, we want to get 

1. patients who Abandoned the ART during the current month
2. Remove patients who Abandoned the ART during the previous month
3. Remove the transferred out ones
4. Remove the suspended 
5. then remove the died 

We should have 5 checks in the composition, but I'm only seeing 4 !

PR #239, opened 2020-03-09
MEPTS-346: fix transferred in query

ynurmahomed:


ningosi (approved):
Changed the transfer states from transfer out to transfer in. Looks fine

PR #244, opened 2020-03-10
Changes in Monthly Report MISAU- A2 (REVIEW)

Edgencio:


api/src/test/java/org/openmrs/module/eptsreports/reporting/intergrated/library/cohorts/ResumoMensalCohortQueriesTest.java
```diff
@@ -85,7 +85,7 @@ public void getDeadPatientsShouldReturn() throws EvaluationException {
     assertTrue(evaluatedCohort.getMemberIds().contains(1023));
   }
 
-  @Test
+  @Ignore
```
cfaife, line 88: why this test is being ignored?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/ResumoMensalCohortQueries.java
```diff
@@ -120,12 +121,12 @@ public CohortDefinition getPatientsWhoInitiatedPreTarvAtAfacilityDuringCurrentMo
     sqlCohortDefinition.addParameter(new Parameter("endDate", "End Date", Date.class));
     sqlCohortDefinition.addParameter(new Parameter("location", "Location", Location.class));
     sqlCohortDefinition.setQuery(
-        ResumoMensalQueries.getPatientsWhoInitiatedPreArtDuringCurrentMonthWithConditions(
+        getPatientsWhoInitiatedPreArtDuringCurrentMonthWithConditions(
```
cfaife, line 124: not a big deal, but at  first look, the method seems to be a part `this` class, and actually not. As we do not  have the habit of  use `this` in the  `project` to refer  instance  variables/methods,  we should call static methods with the class  where it  belongs.

PR #246, opened 2020-03-10
MEPTS-354: Updating TX_TB_Numerator to include checks for Patients with Pulmonary TB Date and Patients on treatment for TB

rubailly:
- Added ` #getPatientsWithObsBetweenDates()` in `TXTBQueries` to get patients with a value coded Obs (Patients with Pulmonary TB Date and Patients marked as Tratamento TB– Inicio (I)) betwen startDate and endDate

- Added `#getPulmonaryTBConceptUuid()` in `HivMetadata` to get `PNEUMONIA, TUBERCULOUS` concept

- Updated `#txTbNumeratorA()` to include a union of `patientswithPulmonaryTbDate` and `patientsWhoInitiatedTbTreatment` in the composition string


ningosi:
@rubailly if you resolve this conflicts, I will be able to give this PR the final pass

api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
```diff
@@ -983,6 +983,13 @@ public Concept getBpostiveConcept() {
     return getConcept(uuid);
   }
 
+  // concept_id = 42
```
ningosi: This concept is also in TbMetadata class. Could we have have it in one place depending on use and relevance?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TXTBCohortQueries.java
```diff
@@ -176,11 +176,35 @@ public CohortDefinition txTbNumeratorA() {
...
+    CohortDefinition patientswithPulmonaryTbDate =
+        TXTBQueries.getPatientsWithObsBetweenDates(
+            "Patients with Pulmonary TB Date",
+            hivMetadata.getPulmonaryTBConceptUuid(),
+            hivMetadata.getPatientFoundYesConcept(),
+            Arrays.asList(hivMetadata.getMasterCardEncounterType()));
+    cd.addSearch(
+        "patientswithPulmonaryTbDate", map(patientswithPulmonaryTbDate, codedObsParameterMapping));
+
+    CohortDefinition patientsWhoInitiatedTbTreatment =
```
ningosi, line 360: This definition looks like it has to exclude something look below
**all patients who started TB Treatment at any point in the previous 6 months before 
to the start date of the reporting period (Numerator for previous 6 months)**

ningosi (commented):
There are section that I have highlighted that need to check

api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
rubailly: [Done](https://github.com/esaude/openmrs-module-eptsreports/pull/246/commits/7a85d96ab02ad6dc0fb86669c517985486291279)

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TXTBCohortQueries.java
rubailly: @ningosi  The dataset doesn't call that method directly, it rather calls `txTbNumerator()` which handles that exclusion.

ningosi:
All the comment having been addressed, the second review has passed and the build is succedded, I will merge this PR to be ready for testing by the BA team

PR #251, opened 2020-03-13
361_c1_fixes

ningosi:
Including both options of the YES/NO for the TB screening for the C1 dataset


psmatsinhe (approved):
Looks good to me

PR #254, opened 2020-03-15
MEPTS-380: Updating logic to retrieve patients who completed treatment

rubailly:
- Updated `NUMBER_ISONIAZID_USAGE_TO_CONSIDER_COMPLETED`to be 5 and `MONTHS_TO_CHECK_FOR_ISONIAZID_USAGE` to be 6 according to the new specs.

- `getMinOrMaxObsDate` can return null  in cas both `seguimentoOrFichaResumoDate` and `fichaClinicaMasterCardDate` sources for end treatment are null. We'll proceed to do the calculations with `calculateNumberOfYesAnswers`

- New specs say to check for 5 yes/continues for 7 months and not 6 months

- Getting the start drugs or end drugs Obs from Ficha de Seguimento (adults and children) or Ficha Resumo **OR** The start drugs or end drugs Obs from Ficha Clinica-MasterCard is handled by  `getMinOrMaxObsDate`


api/src/main/java/org/openmrs/module/eptsreports/metadata/CommonMetadata.java
```diff
@@ -50,6 +50,13 @@ public Concept getYesConcept() {
     return getConcept(uuid);
   }
 
+  // concept_id = 1257
```
psmatsinhe: This already exist

psmatsinhe (approved):

PR #263, opened 2020-03-16
Mepts 374 echo database

ningosi:
Sorted out differences resulting on the viral load numerator and denominator in the PVLS report


ynurmahomed:
@ningosi could you briefly explain the issue and the fix here?

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/ViralLoadQueries.java
```diff
@@ -61,7 +71,7 @@ public static String getPatientsWithViralLoadSuppression(
         vlConceptQuestion,
         vlQualitativeConceptQuestion,
         mastercardEncounter,
-        vlConceptQuestion);
+        vlConceptQuestion, vlConceptQuestion, vlConceptQuestion,vlQualitativeConceptQuestion);
```
ynurmahomed, line 74: @ningosi forgot to say that we should migrate away from `String.format`

PR #270, opened 2020-03-18
Mepts 384 when pregnant and breastfeeding record pregnant

ningosi:
When the pregnancy and breastfeeding state apperas on the same date, we need to record that state as pregnancy


psmatsinhe (approved):
Looks good to me

PR #274, opened 2020-03-18
MEPTS-386: Added #getFileOpeningObs metod to handle Ficha Resumo Encounter Type

rubailly:
- Added `#getFileOpeningObs` method to process `fichaResumoEncounterType`

- Added  `CalculationResultMap fileOpeningObsMap` in both `BreastfeedingDateCalculation` and `PregnantDateCalculation` to handle `fichaResumoEncounterType` separately from `CalculationResultMap lactatingMap` and `CalculationResultMap pregnantMap`

- Updated `#getResultantDate` in `BreastfeedingDateCalculation` and `#getRequiredDate` in `PregnantDateCalculation` to accept `fileOpeningObsMap` as a parametter.

- Updated the above methods to get the list of file opening date obs as:
`      ListResult fileOpeningResults = (ListResult) fileOpeningObsMap.get(pId);
      List<Obs> fileOpeningObs = EptsCalculationUtils.extractResultValues(fileOpeningResults);`

**TODO**: 

- Replace ` this.isLactating(lastVlDate, fileOpeningObs),` in `BreastfeedingDateCalculation` with a dedicated method to handle `fichaResumoEncounterType`

- Replace `isPregnantDate(lastVlDate, fileOpeningObsList),`in `PregnantDateCalculation` with a dedicated method to handle `fichaResumoEncounterType`

### DIFF
--- a/api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/common/EPTSCalculationService.java
+++ b/api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/common/EPTSCalculationService.java
@@ -29,6 +29,7 @@ import org.openmrs.module.eptsreports.reporting.utils.EptsCalculationUtils;
 import org.openmrs.module.reporting.common.TimeQualifier;
 import org.openmrs.module.reporting.data.patient.definition.EncountersForPatientDataDefinition;
 import org.openmrs.module.reporting.data.patient.definition.ProgramEnrollmentsForPatientDataDefinition;
+import org.openmrs.module.reporting.data.patient.definition.SqlPatientDataDefinition;
 import org.openmrs.module.reporting.data.person.definition.ObsForPersonDataDefinition;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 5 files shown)


ningosi:
There are conflicts in the config file

api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/common/EPTSCalculationService.java
```diff
@@ -79,6 +80,65 @@ public CalculationResultMap getObs(
...
+   * @param fileOpeningConcept
+   * @param pregnantOrLactatingConcept
+   * @param yesConcept
+   * @param cohort
+   * @param location
+   * @param startDate
+   * @param context
+   * @return
+   */
+  public CalculationResultMap getFileOpeningObs(
```
ningosi, line 96: if you pass those parameters in the getObs method, you will get an entire obs object that you will out that date for your comparison. What am saying we can reuse already existing

rubailly: @ningosi the problem is `getObs` is using `setOnOrBefore` and `setOnOrAfter`.
With the new specs, we want to use the **value datetime** of the obs. Not when the actual obs was created.

rubailly: @ningosi @psmatsinhe to clarify what I'm saying, the new specs say that if the encounter type is 53, we first check if there is pregnancy obs on that encounter with a value coded 'Yes'.
We then check on that same encounter if there is a file opening date obs with a value date-time within the reporting range.
I might be missing something as I don't really have a deep understanding of the PVLS report specs, but I couldn't see how the `getObs` method was going to satisfy the above requirements.

ningosi: @rubailly the two parameters setOnOrBefore and setOnOrAfter if provided, they will only help you get results to that limit, for example we need results that will NOT after endDate, therefore, those parameters will NOT hinder you from getting the right obs.  as you can see here https://github.com/esaude/openmrs-module-eptsreports/blob/master/api/src/main/java/org/openmrs/module/eptsreports/reporting/calculation/common/EPTSCalculationService.java#L70-L72, only the endDate is applied, startDate is optional

ningosi: for checking yes(1065) as the coded obs, still that method could cover, let me go through the specs again

ningosi:
Closing this PR, will provide a new one

PR #281, opened 2020-03-20
Mepts 386 pregnant vs breastfeeding

ningosi:
There was a need to filter out female patients who have encounter type 53 and their value date time of the file opening concept matches the criteria for pregnant and breastfeeding


rubailly:
LGTM

PR #295, opened 2020-03-30
MEPTS-389 Tx-TB Fix in Base Query Patients

iteliolucas:


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TXTBCohortQueries.java
```diff
@@ -266,29 +266,32 @@ public CohortDefinition codedYesTbScreening() {
     return cd;
   }
 
+  /*
+  * Patients who started art on period considering the transferred in for that same period
+  * and patients who started art before period also considering the transferred in for that same period
+  */
```
cfaife, line 272: how  parameters(date period and  location) are passed here?

cfaife (changes requested):

PR #296, opened 2020-03-31
Mepts 398 rm differences in certification of tx ml

cfaife:


api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TxMlQueries.java
```diff
@@ -25,106 +25,90 @@ public static String getPatientsWhoMissedAppointment(
...
-            + ")all_patients "
-            + " WHERE "
-            + "	DATE_ADD(return_date, INTERVAL 28 DAY)  >= DATE_ADD(:startDate, INTERVAL -1 DAY) "
-            + "	AND DATE_ADD(return_date, INTERVAL 28 DAY) < :endDate";
+        " SELECT final.patient_id    "
+            + "             from(    "
+            + "                 SELECT    "
+            + "                     most_recent.patient_id, Max(most_recent.value_datetime) final_encounter_date    "
+            + "                 FROM   (SELECT fila.patient_id, o.value_datetime from (   "
+            + "                             SELECT enc.patient_id,    "
```
psmatsinhe, line 33: @cfaife this query is not behaving the way you think it should, you doing a select with one table and group by with other table, you need to use the same thing!

> Select pa.patient_id instead of enc.patient_id

psmatsinhe (changes requested):

psmatsinhe (approved):
Looks good to me

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/queries/TxMlQueries.java
psmatsinhe: @cfaife Was having problem with location id

PR #301, opened 2020-04-01
MEPTS-373: Changes in Tx-TB Denominator - New Disaggregations

rubailly:
- Most of the cohort queries are in `TXTBCohortQueries`

- `#getPatientsWhoHaveSentSpecimen` method is used for **BR-8 SPECIMEN SENT**

- `#getPatientsWhoHaveGeneXpert` method is used for **BR-9 DIAGNOSTIC TEST – GENXPERT MTB/RIF**

- `#getSmearMicroscopyOnly` method is used for **BR-10 DIAGNOSTIC TEST – SMEAR MICROSCOPY ONNLY**

- `#getAdditionalTest` method is used for **BR-11 DIAGNOSTIC TEST – ADDIONAL TEST**

- `#getPositiveResultsReturned` method is used for **BR-12 POSITIVE RESULTS RETURNED**

- They all call `#getPatientsWithCodedObsBetweenDates` which uses reporting `CodedObsCohortDefinition` to get an obs between dates.


api/src/main/java/org/openmrs/module/eptsreports/metadata/HivMetadata.java
```diff
@@ -419,6 +419,14 @@ public Concept getCD4PercentConcept() {
     return getConcept(uuid);
   }
 
+  // concept_id=307
```
ningosi: Any Tb related metadata that is only used in TB queries should be created in TbMetadata.java, if such metadata are shared across the application, we would create those in CommonMetadata.java

api/src/main/java/org/openmrs/module/eptsreports/reporting/library/cohorts/TXTBCohortQueries.java
```diff
@@ -356,4 +356,44 @@ public CohortDefinition getNewOnArt() {
...
+  }
+
+  /**
+   * BR-8 Specimen Sent Get patients from denominator AND tb_screened AND specimen_sent
+   *
+   * @return CohortDefinition
+   */
+  public CohortDefinition specimenSent() {
+    CompositionCohortDefinition definition = new CompositionCohortDefinition();
+    definition.setName("specimenSent()");
```
ningosi, line 718: Maybe a descriptive name here would serve better than using the method name

ningosi (commented):
Code looks fine, except of putting certain variables in the wrong place. I will wait for the run results then approve it

PR #309, opened 2020-04-13
Mepts 424 (#308)

rubailly:
* MEPTS-424: Added a check to consider days when calculating difference between ART and IPT start dates

* MEPTS-424: Calculation should check if earliest isoniazid start date is in the previous 6 months

* MEPTS-434: Started ART query shoul check based on endDate of the current reporting period. See MEPTS-424 for details

* MEPTS-434: NewlyOrPreviouslyEnrolledOnARTCalculation should use the actual endDate parameter when querying for patients who start ART

Co-authored-by: Bailly RURANGIRWA <rubailly@gmail.com>


psmatsinhe (approved):
Looks good to me

PR #315, opened 2021-02-18
Mepts 765 mq report cat 7 change request

helderjosue:
Changes on CompositionStrings for MQ Cat 7


helderjosue:
> Changes on CompositionStrings for MQ Cat 7

